PR #3, opened 2023-05-22
V14 update

### LINKED ISSUE #2: Setup Test Fixtures
- [x] Workstations
- [x] Suppliers
- [x] Item Groups (for consistency with ERPNext demo)
- [x] Items
- [x] Warehouses
- [x] BOMs
- [x] Production Plan / Material Requests / Purchase Receipt
- ~~[ ] Asset(s) (deffered~~

### LINKED ISSUE #4: Add Typing for scan APIs and functions

### LINKED ISSUE #6: Remove APIs for Job Cards
These were specific to a customer workflow that does not have a generalize implementation pattern

### LINKED ISSUE #9: Restructure decision matrix
Instead of the confusing tuple-keyed dictionary, let's use another layer of nesting:
```python
"Item" : {
    "Purchase Receipt": [
	("add_or_increment", "Purchase Receipt Item", "item_code", target.item_code, target),
    ]
```

This will allow for clearer access of the correct function(s) and to make it hookable in the future.

### LINKED ISSUE #10: Remove Asset-related scanning features
To be added back in the future with a more complete design that incorporates more features of the Asset module, eg Asset Maintenance.

agritheory:
- [x] Workstations
- [x] Suppliers
- [x] Item Groups (for consistency with ERPNext demo)
- [x] Items
- [x] Warehouses
- [x] BOMs


beam/beam/scan/__init__.py
```diff
@@ -40,7 +40,7 @@ def get_handling_unit(handling_unit: str) -> frappe._dict:
 		filters={"handling_unit": handling_unit},
 		fields=[
 			"item_code",
-			"actual_qty",
+			"SUM(actual_qty) as actual_qty",
```
agritheory: @Alchez Why? This should only return the most recent handling unit which is why there are the `order_by` and `limit` arguments below

Alchez: Oh, that was actually the [original SQL query](https://github.com/agritheory/beam/pull/3/commits/5c0824af9950473588f2f82bd9daa2fb4237e01d#diff-7082f090171a8a5622ce14e7d742696adc0d06e4df88dbfa976c29e8bb70d1e6R38) and the last commit was because I had forgotten to switch it back to the `sum` format. I added the `group_by` param too, because the sum operation would need it.

beam/beam/scan/__init__.py
```diff
@@ -0,0 +1,353 @@
...
+			],
+			"Packing Slip": [
+				("filter", "Packing Slip Item", "warehouse", target),
+			],
+			"Purchase Invoice": [
+				("filter", "Purchase Invoice Item", "warehouse", target),
+			],
+			"Purchase Receipt": [
+				("filter", "Accepted Warehouse (Purchase Receipt Item)", target),
+			],
```
Alchez: @agritheory, this one seems incorrect since the tuple only has 3 elements. First, should we point this to the "Accepted Warehouse" fieldname as the third param. Second, should the tuple be changed to a dict or namedtuple?

agritheory: I'd vote for `frappe._dict`, and yes, this is wrong

PR #14, opened 2023-05-26
Draft: Print server integration

### LINKED ISSUE #13: Printer Server integrations
Print Handling Unit APIs and examples

I think that usage of this feature is going to be very specific to each business, so I'm not sure how to deliver all the functionality. Might be worth a configuration page, potentially per-user.

agritheory:


beam/beam/printing.py
```diff
@@ -0,0 +1,104 @@
...
+	if isinstance(doc, str):
+		doc = frappe._dict(json.loads(doc))
+	if not print_format:
+		print_format = frappe.get_meta(doctype).get("default_print_format")
+	print_format = frappe.get_doc("Print Format", print_format)
+	try:
+		cups.setServer(print_settings.server_ip)
+		cups.setPort(print_settings.port)
+		conn = cups.Connection()
+		if print_format.raw_printing == 1:
```
Alchez, line 39: Should this entire `if...else` block be nested under an `if not file_path` condition? Because your `output` variable doesn't seem to be doing anything if the user has provided a file path.

agritheory: It's actually an indentation error (fixed)

PR #18, opened 2023-06-01
feat: handle scanning of non-handling units

### LINKED ISSUE #5: Harmonize Form return values with the `get_item_details`
By also calling `get_item_details`, we can return more complete results from a barcode (especially a Handling Unit) scanned and then update the UI correctly

### LINKED ISSUE #7: Change pre-process strategy
It seems that the approach for fetching handling units' details is relatively complete. We need to expand the strategy for other Beam-enabled data types: Item, Asset and Warehouse

The Item's return value should be similar to what is done with ERPNext's `get_item_details` function. An asset would be similar but not enabled in all form inputs.

Warehouse might be the trickiest because we want actions with it to be smart:
For example in a Stock Entry:
- Scan Item (populates Stock Entry Detail row)
- Scan Warehouse (updates the warehouse is the row that was just added)

Or:
- Scan Warehouse (no rows; populates an Item if it is the only item in that warehouse, otherwise  it only populates the warehouse)

Additional design requirements will be added to this ticket and the project documentation over time

### LINKED ISSUE #16: Warehouse scanning
Refining requirements about scanning in Stock Entry:
Both scanning of an Item and Handling Unit should result in the same behaviors as with other interfaces with the following caveats:
- A Handling Unit's warehouse should be set as the source warehouse
- The Item's default warehouse should be set as the source warehouse

When scanning a warehouse in a Stock Entry, it will always be treated as the target warehouse.

Alchez:
Fixes #7

### DIFF
--- a/beam/beam/scan/__init__.py
+++ b/beam/beam/scan/__init__.py
@@ -3,6 +3,7 @@ import json
 from typing import Any, Optional, Union
 
 import frappe
+from erpnext.stock.doctype.stock_entry.stock_entry import StockEntry
 from erpnext.stock.get_item_details import get_item_details
 
 
@@ -255,12 +256,7 @@ def get_form_action(barcode_doc: frappe._dict, context: frappe._dict) -> list[di
 	if barcode_doc.doc.doctype == "Handling Unit":
 		target = get_handling_unit(barcode_doc.doc.name)
 	elif barcode_doc.doc.doctype == "Item":
-		if context.frm == "Stock Entry":
-			# TODO: ERPNext internally uses the `get_item_details` function in the Stock Entry controller
-			# based on the SE's purpose; figure out a way to use that (maybe fetch current doc to call,
-			# or write a custom method)
-			pass
-		else:
+		if context.frm != "Stock Entry":
 			# the base `get_item_details` cannot handle doctypes whose items table name doesn't have
 			# "Item" in it, which will fail for Stock Entry
 			target = get_item_details(
@@ -271,6 +267,16 @@ def get_form_action(barcode_doc: frappe._dict, context: frappe._dict) -> list[di
 					"currency": frappe.defaults.get_user_default("Currency"),
 				}
 			)
+		else:
+			stock_entry = StockEntry(frappe._dict(context.doc))
+			if not stock_entry.stock_entry_type:
+				stock_entry.purpose = "Material Transfer"
+				stock_entry.set_stock_entry_type()
+			target = stock_entry.get_item_details({"item_code": barcode_doc.doc.name})
+			target.item_code = barcode_doc.doc.name
+			target.barcode = barcode_doc.barcode
+			# only required for first scan, since quantity by default is zero
+			target.qty = 1
 
 	if not target:
 		return []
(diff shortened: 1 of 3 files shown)


beam/public/js/scan/scan.js
```diff
@@ -238,25 +238,25 @@ class ScanHandler {
...
 				)
 			})
 		) {
 			if (!cur_frm.doc.items.length || !cur_frm.doc.items[0].item_code) {
 				cur_frm.doc.items = []
 			}
-			let row = cur_frm.add_child('items', barcode_context)
-			frappe.ui.form.handlers[cur_frm.fields_dict.items.df.options].uom.map(r => {
+			let row = cur_frm.add_child('items', barcode_context.context)
+			frappe.ui.form.handlers[cur_frm.fields_dict.items.df.options].uom?.map(r => {
```
agritheory: I think we want to remove this line

agritheory (changes requested):
Maybe worth having a conversation about best practices for setting rows and individual values in child tables. I have conflicting opinions.

beam/public/js/scan/scan.js
Alchez: I've removed that particular line, but we have a couple more instances that it's used. Should we remove those as well?

https://github.com/agritheory/beam/blob/54f6766ca9d7c6f42bac78573896370b70ff47b6/beam/public/js/scan/scan.js#L130-L134

https://github.com/agritheory/beam/blob/54f6766ca9d7c6f42bac78573896370b70ff47b6/beam/public/js/scan/scan.js#L177-L181

beam/beam/scan/__init__.py
```diff
@@ -1,345 +1,519 @@
...
+def get_form_action(barcode_doc: frappe._dict, context: frappe._dict) -> list[dict[str, Any]]:
+	target = None
+	if barcode_doc.doc.doctype == "Handling Unit":
+		target = get_handling_unit(barcode_doc.doc.name)
+	elif barcode_doc.doc.doctype == "Item":
+		if context.frm == "Stock Entry":
+			# TODO: ERPNext internally uses the `get_item_details` function in the Stock Entry controller
+			# based on the SE's purpose; figure out a way to use that (maybe fetch current doc to call,
+			# or write a custom method)
+			pass
```
Alchez: @agritheory I've added a comment for how this could be handled, but I'm not sure about the implementation detail you prefer

beam/public/js/scan/scan.js
agritheory: Yes please, I think they're also no longer required.

beam/beam/scan/__init__.py
agritheory: Let's match what ERPNext is doing as closely as we can. It probably makes sense to pass the stock entry type in the form context.

Alchez: Should we enforce having the user set a purpose before they scan a barcode/HU then?

agritheory: Let's use a fallback of 'Material Transfer' instead; not correct, just least wrong.

Alchez:
@agritheory, this one is ready to be reviewed

beam/public/js/scan/scan.js
```diff
@@ -127,11 +128,6 @@ class ScanHandler {
 					row.source_warehouse = field.context.warehouse
 					if (cur_frm.doc.operation && cur_frm.doc.operation.includes('Mixing - ')) row.required_qty = row.stock_qty = 0
```
agritheory: Let's remove this reference to Operation - this is customer specific code

PR #33, opened 2023-07-01
Make action configuration hookable

### LINKED ISSUE #29: Make action configuration hookable
Provide an API to override the action as addressed in the decision matrix:

For example, [when an Item barcode is scanned in the Delivery Note list view](https://github.com/agritheory/beam/blob/version-14/beam/beam/scan/__init__.py#L156), a custom JS function could be called instead.

This means extending the API to two places: one that calls `frappe.get_hooks` and collects any configurations that are required and [a second that calls a function](https://github.com/agritheory/beam/blob/version-14/beam/public/js/scan/scan.js#L97) (if the key `function` exists in the return object). 

This JS code should be refactored to allow different actions to specified per return value - something that is [handled individually inside each class method](https://github.com/agritheory/beam/blob/version-14/beam/public/js/scan/scan.js#L114-L119). I don't have a great use case for this but I suspect that somebody will want it.

agritheory:
This PR adds the ability to extend Beam with custom hooks. Closes #29

~~The monkey patch in the tests does not work and I'm not sure how to solve it. Marking as draft until it's resolved.~~ 

I figured out the monkey patch problem; I had actually solved in another project.

Client hooks, configuration and documentation also added.


agritheory:
@fproldan Please review when you get a chance

beam/hooks.py
```diff
@@ -217,3 +221,40 @@
...
+			},
+		]
+	}
+}
+
+beam_listview = {
+	"Item": {
+		"Item": [
+			{
+				"action": "show_message",
```
fproldan: Due to this change we need to update [this test ](https://github.com/agritheory/beam/blob/beam_hooks/beam/tests/test_scan.py#L33). Now the action is `show_message` instead of `route`, As the hooked config took place over the app configuration as should be.

beam/docs/hooks.md
```diff
@@ -0,0 +1,46 @@
...
+
+```python
+# hooks.py
+
+beam_listview = {
+	"Item": {
+		"Delivery Note": [
+			{"action": "filter", "doctype": "Delivery Note Item", "field": "item_code"},
+			{"action": "filter", "doctype": "Packed Item", "field": "item_code"}
+		],
```
fproldan, line 17: very minor, missing `}` here

beam/hooks.py
agritheory: Glad you caught this. This code should be removed from this file, I was just using it to test the functionality. Normally it would be provided in the customer's app, not part of Beam.

beam/beam/scan/__init__.py
```diff
@@ -87,6 +87,7 @@ def get_list_action(barcode_doc: frappe._dict, context: frappe._dict) -> list[di
 	target = barcode_doc.doc.name
 	if barcode_doc.doc.doctype == "Handling Unit":
 		if barcode_doc.doc.get("parenttype") == "Packing Slip":
+			# TODO: is this check correct and/or required?
```
fproldan, line 90: Perhaps adding a test case will help us answer the question, I'm not quite sure.

beam/beam/scan/config.py
```diff
@@ -1,32 +1,39 @@
...
-			'Stock Entry',
-			'Stock Reconciliation',
-		],
+	scannable_doctypes = set()
+	listview_doctypes = set()
+	frm_doctypes = set()
+	hooked_listview = frappe.get_hooks("beam_listview")
+	hooked_frm = frappe.get_hooks("beam_frm")
+	beam_client = frappe.get_hooks("beam_client")
+	print(beam_client)
```
fproldan: a print slipped in here

beam/beam/scan/__init__.py
agritheory: I'll add a test case and docs for Packing Slip but it's outside of the scope of this feature/ PR.

fproldan (commented):
A couple of minor comments. LGTM, a very nice feature to have!

agritheory (commented):
@fproldan I fixed everything you mentioned except for the comment about the Packing Slip which I'll tackle in a future PR.

PR #45, opened 2023-07-11
fix: ignore permissions on barcode creation

### LINKED ISSUE #44: Non-admin users encounter permissions error
image

This likely needs an ignore_permissions flag in `create_handling_unit`

Possibly related, the default permissions for Handling Unit are too restrictive, Stock User should have complete permissions.

agritheory:


agritheory:
Resolves #44

PR #46, opened 2023-07-11
docs: document features - first pass

agritheory:


HKuz:
Latest commits add screen shots and update text for the docs. I came across two potential issues while trying to create screen shots:

- In the Item listview, I simulated scanning a HU for a fruit, and got Page Not Found error. The route was `app/item/MAT-PRE-2023-00001` (the purchase receipt name the item was included in where it should have been the item name)
- When I created Stock Entries of type `Material Transfer for Manufacture` (off a work order) and a plain `Material Transfer`, the stock ledger entries didn't have the handling unit attached for any of the items moved. (Screen shot shows one example). I thought this would be automatic - if there's some config I may have missed, I'll make sure to add that to the docs. I'd like to have one screen shot of a stock entry example in there.

Screenshot 2023-07-17 at 6 12 41 PM

beam/docs/handling_unit.md
```diff
@@ -72,10 +124,15 @@ In both these cases, there is no offsetting movement or creation of items.
...
 
-## Material Receipt
+### Material Receipt
 In the case of Material Receipt, a new Handling Unit is generated for each item. 
 
 | Item             | Warehouse          | Handling Unit  |       Quantity |
 | ---------------- | ------------------ | -------------- | --------------:|
 | Cocoplum         | Storeroom          |            123 |          20 Ea |
 
+## Warehouses
```
agritheory: @HKuz This should get moved out of this file, the barcode on a Warehouse is just a barcode, same as on an Item and it is definitely not a Handling Unit. I say we move it to index.md along with the explanation of how barcodes are auto generated for Item and Warehouse.

agritheory:
@HKuz 

> In the Item listview, I simulated scanning a HU for a fruit, and got Page Not Found error. The route was app/item/MAT-PRE-2023-00001 (the purchase receipt name the item was included in where it should have been the item name)

I think we need to make the change to something like this:
```python
# 243 scan/__init__.py
"Item": [{"action": "route", "doctype": "Item", "field": "Item", "target": "target.item_code"}],
```

> When I created Stock Entries of type Material Transfer for Manufacture (off a work order) and a plain Material Transfer, the stock ledger entries didn't have the handling unit attached for any of the items moved. (Screen shot shows one example). I thought this would be automatic - if there's some config I may have missed, I'll make sure to add that to the docs. I'd like to have one screen shot of a stock entry example in there.

I think this is a bug. Let's check the tests to confirm. 

I think we need to move the Warehouses section out of Handling Unit and into index.md. 
This looks great, it feels so much more polished.

Issue #42, opened 2023-07-11
Add validation to disallow overconsumption of qty in a handling unit


agritheory:
Assume the Handling Unit has a qty of 40. The follow transaction should fail:

| Item             | Warehouse          | Handling Unit  |       Quantity |
| ---------------- | ------------------ | -------------- | --------------:|
| Cocoplum         | Storeroom          |            123 |          41 Ea |

- [x] Add validation function that throws an error when this is the case
- [x] Add test case that confirms that an error is being raised (`@pytest.mark.xfail()`)

agritheory:
Reopening as this current implementation is a little off the mark. It's passes the specific xfail test, but also fails a bunch of other tests, conflicting with net quantity features.

Update: we want to throw a negative inventory error here, but have it specific to the handling unit. [There is some code](https://github.com/frappe/erpnext/blob/7ab55b1bb293517aaac11193f25efdf6ccbc7bb4/erpnext/stock/doctype/serial_and_batch_bundle/serial_and_batch_bundle.py#L193) in the ERPNext repo to base this on - we need to detect if the transaction is incoming or outgoing and switch the greater than/ less than operator appropriately.

PR #48, opened 2023-07-19
Add validation to disallow overconsumption of qty in a handling unit

fproldan:
#42

### DIFF
--- a/beam/beam/handling_unit.py
+++ b/beam/beam/handling_unit.py
@@ -68,13 +68,18 @@ def validate_handling_unit_overconsumption(doc, method=None):
 	):
 		return doc
 
+	if doc.doctype == "Stock Entry":
+		qty_field = "actual_qty"
+	else:
+		qty_field = "stock_qty"
+
 	for row in doc.get("items"):
 		if not hasattr(row, "handling_unit") or not row.handling_unit:
 			continue
 
 		hu = get_handling_unit(row.handling_unit)
 
-		if hu.stock_qty < abs(row.qty):
+		if abs(hu.stock_qty - row.stock_qty) < frappe.get_precision(row.doctype, qty_field):
 			frappe.throw(
 				_(f"Row #{row.idx}: the Handling Unit for Item {row.item_code} has qty of {hu.stock_qty}.")
 			)
--- a/beam/beam/handling_unit.py
+++ b/beam/beam/handling_unit.py
@@ -50,8 +50,11 @@ def generate_handling_units(doc, method=None):
 
 
 @frappe.whitelist()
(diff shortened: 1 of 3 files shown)


beam/beam/handling_unit.py
```diff
@@ -44,3 +47,33 @@ def generate_handling_units(doc, method=None):
 		row.handling_unit = handling_unit.name
 
 	return doc
+
+
+@frappe.whitelist()
+def validate_handling_units_overconsumption(doc, method=None):
```
agritheory: Can we change the name of this to `validate_handling_unit_overconsumption` - it reads a tiny bit better.

agritheory (changes requested):
A late addition to the requirements for this ticket, but I think we need to cover the Purchase Return workflow also, even though it's very much an edge case here. I didn't even think of it until I was reviewing. https://docs.erpnext.com/docs/v13/user/manual/en/stock/purchase-return

beam/beam/handling_unit.py
```diff
@@ -71,7 +74,7 @@ def validate_handling_units_overconsumption(doc, method=None):
 
 		hu = get_handling_unit(row.handling_unit)
 
-		if hu.stock_qty < row.qty:
+		if hu.stock_qty < abs(row.qty):
```
agritheory:
```
frappe.get_precision("Purchase Order Item", "stock_qty")
abs(hu.stock_qty- row.stock_qty) < precision_places
```

beam/beam/handling_unit.py
```diff
@@ -68,13 +68,18 @@ def validate_handling_unit_overconsumption(doc, method=None):
 	):
 		return doc
 
+	if doc.doctype == "Stock Entry":
```
agritheory, line 71: This would make for a nice ternary: 
```python
qty_field = "actual_qty" if doc.doctype == "Stock Entry" else "stock_qty"
```

beam/beam/handling_unit.py
```diff
@@ -68,13 +68,18 @@ def validate_handling_unit_overconsumption(doc, method=None):
...
+		qty_field = "stock_qty"
+
 	for row in doc.get("items"):
 		if not hasattr(row, "handling_unit") or not row.handling_unit:
 			continue
 
 		hu = get_handling_unit(row.handling_unit)
 
-		if hu.stock_qty < abs(row.qty):
+		if abs(hu.stock_qty - row.stock_qty) < frappe.get_precision(row.doctype, qty_field):
```
agritheory, line 82: Perfect.

agritheory (approved):

Issue #12, opened 2023-05-26
Traceability Reporting


agritheory:
We need to test the life cycle of a handling unit. Populating this data also gives use the traceability API. 

Purchase Receipt or Purchase Invoice with "Update Stock", thru Manufacturing cycle (2x) to Sale / disposition
```mermaid
flowchart TD
	Z[Purchase Invoice] --> C[Work Order]
	A[Purchase Receipt] --> C[Work Order]
	C --> D[Stock Entry - Transfer for Manufacture] 
	D --> E[Stock Entry - Manufacture]
	E --> C
        E --> F[Delivery Note]
        E --> G[Sales Invoice]
```

agritheory:
- Should require a filter for at least one of Handling Unit, Delivery Note or Sales Invoice with update stock checked
- Let's use the Query Builder so that this is Postgres compatible
- Add documentation and a link in index.md
- Script the submission of documents so that we can have a test for this

agritheory:
@HKuz Please add some docs for this once we are ready to merge

PR #50, opened 2023-07-21
Traceability Reporting

fproldan:
#12 

Captura de pantalla 2023-07-26 a la(s) 09 12 57


federicocalvo:
the ideas to visualize it in a correct way started from here, then it was done what seemed more feasible, but anything can be deepened or spend more time to achieve any of these results.

Option1:

Captura de pantalla 2023-07-21 a la(s) 14 15 23

Option2:

Captura de pantalla 2023-07-21 a la(s) 14 14 21

agritheory:
@federicocalvo @fproldan These are both great options, I like where we are headed. We can ask the customer who requested this for what would would best for them. 

@Rgottsch1 Can you look though the screenshots in this Pull Request and give some feedback.

Rgottsch1:
Both are very good, I like the presentation in Option 1 slightly better.  Great work!

Issue #49, opened 2023-07-19
Handling Unit not tracking with certain material transfer-type stock entries


HKuz:
Creating Stock Entries of type `Material Transfer for Manufacture` (off a work order) and a plain `Material Transfer`, the stock ledger entries didn't show the handling unit for any of the items moved. (Screen shot shows one example). Both stock entries moved quantities less than the full handling unit qty.

Expected behavior: 
- when the qty moved is less than what's in the HU, the original HU stays in the source WH and shows a reduction, and a new HU is created for the items moved to the target WH
- when the quantity needed is for everything in the HU, the HU 'travels' to the new WH

<img width="1211" alt="Screenshot 2023-07-17 at 6 12 41 PM" src="https://github.com/agritheory/beam/assets/9591826/a547d17d-5b95-4ad7-a1cf-3a8997b4b97d">

HKuz:
@fproldan @agritheory - in testing my code changes for this issue, I made the following tweaks to the over-consumption validation code below. This checks that the doc's row qty is not larger than the qty in the HU, within "precision" decimal accuracy:

```py
        if doc.doctype == "Stock Entry":
		# qty_field = "actual_qty"  # Changed to transfer_qty EDIT: actual_qty is total qty available for item in given WH (checks source WH, if not defined uses target WH)
		qty_field = "transfer_qty"
	else:
		qty_field = "stock_qty"

        # [...  other code in for loop]
                #  New error check:
                precision = int(frappe.get_precision(row.doctype, qty_field))
		if row.get(qty_field) - hu.stock_qty > 1 / (10**precision):
			frappe.throw(
				_(f"Row #{row.idx}: the Handling Unit for Item {row.item_code} has qty of {hu.stock_qty}.")
			)
```

Here's the example in inventory_tools that checks for float equality (within "precision" decimal accuracy). In this case, the precision was 3, so the absolute difference can't be more than 0.001.

```py
precision = int(frappe.get_precision("Purchase Order Item", "stock_qty"))
diff = abs(items_fg_qty - subc_fg_qty)
if diff > (1 / (10**precision)):
    # ...
```

I'll add human-readable comment explanations in both areas.

agritheory:
This problem is in scope of the larger net quantity issue. We will want a test that addresses this case specifically to make sure there's no regression.

agritheory:
Resolved with #63

PR #51, opened 2023-07-27
fix: add manufacturing transfers to handling unit

HKuz:
Addresses #49 

This bug was partially due to user error - the handling unit only attaches to an item when the item is 'scanned' or is manually linked in the Edit form for the item. This PR also adds code for when a HU is present, if the qty being moved is less than the HU's stock_qty (within qty field's precision number of decimals accuracy), it generates a new HU for the transferred qty. There are a few tweaks to the over-consumption validation - the main one is to use `transfer_qty` field for Stock Entries - digging into the code, `actual_qty` is what is available at the source warehouse as of the latest SLE and may be much more than what's being moved. (E.g - a 'Material Transfer for Manufacture' moves a 1lb of butter out of the Refrigerator for pie filling, but `actual_qty` is the 50lbs that's available in the fridge).

I also commented a conversion in `get_handling_unit` - it was converting parchment paper back to Box units (vs. stock UOM Nos). The query is summing `actual_qty` off of SLEs for the `stock_qty` field, so it's already in stock_uoms.

For testing, I checked the following scenarios and all worked as expected:
- Transferred half the required materials for Gooseberry Pie Filling in 'Material Transfer for Manufacture' SE
- Transferred the other half of required materials for Gooseberry Pie Filling in 'Material Transfer for Manufacture' SE
- Created all Gooseberry Pie Filling in 'Manufacture' SE
- Moved all Pie Crust required materials in 'Material Transfer for Manufacture' SE, create all pie crusts in 'Manufacture' SE
- Moved filling/crusts and created Gooseberry Pie
- Created Delivery Note for Gooseberry Pie

### DIFF
--- a/beam/beam/scan/__init__.py
+++ b/beam/beam/scan/__init__.py
@@ -79,7 +79,7 @@ def get_handling_unit(handling_unit: str) -> frappe._dict:
 
 	if _sle:
 		sle.update({**_sle})
-		sle.stock_qty = sle.stock_qty / sle.conversion_factor
+		# sle.stock_qty = sle.stock_qty / sle.conversion_factor
 
 	sle.conversion_factor = frappe.get_value(
 		"UOM Conversion Detail",
--- a/beam/beam/scan/__init__.py
+++ b/beam/beam/scan/__init__.py
@@ -79,7 +79,7 @@ def get_handling_unit(handling_unit: str) -> frappe._dict:
 
 	if _sle:
 		sle.update({**_sle})
-		# sle.stock_qty = sle.stock_qty / sle.conversion_factor
+		sle.qty = sle.stock_qty / sle.conversion_factor
 
 	sle.conversion_factor = frappe.get_value(
 		"UOM Conversion Detail",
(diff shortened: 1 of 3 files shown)


beam/beam/scan/__init__.py
```diff
@@ -79,7 +79,7 @@ def get_handling_unit(handling_unit: str) -> frappe._dict:
 
 	if _sle:
 		sle.update({**_sle})
-		sle.stock_qty = sle.stock_qty / sle.conversion_factor
```
agritheory, line 82: This really points out an API/ contract problem. This field should be called qty and be returned as part of the `get_handling_unit` function

agritheory:
> For testing, I checked the following scenarios and all worked as expected:

> Transferred half the required materials for Gooseberry Pie Filling in 'Material Transfer for Manufacture' SE
> Transferred the other half of required materials for Gooseberry Pie Filling in 'Material Transfer for Manufacture' SE
> Created all Gooseberry Pie Filling in 'Manufacture' SE
> Moved all Pie Crust required materials in 'Material Transfer for Manufacture' SE, create all pie crusts in 'Manufacture' SE
> Moved filling/crusts and created Gooseberry Pie
> Created Delivery Note for Gooseberry Pie

This test case or an equivalent one that tests correcting netting of quantities should be in the test suite. The specifics are in conflict with the use case that Federico brought up the other day.  I'll open a ticket related to that separately.

PR #55, opened 2023-08-12
ci: add test workflow, migrate to Python semantic release

### LINKED ISSUE #32: Run test suite in CI

HKuz:
Addresses #32

Adds CI test workflow and migrates to Python semantic release. This incorporates the db logger patch so tests will run on the latest frappe/erpnext versions. That said, some tests are failing - at least one the reason is fixed by some code tweaks in PR #51 .


agritheory:
@HKuz Let's merge this. I will make another PR that refactors the overconsumption validation, which is what is causing these to fail.

PR #62, opened 2023-08-24
docs: add handling unit traceability report

HKuz:
Includes the report documentation (against #12 ), also minor tweak to conform item spelling in the test data.


agritheory (changes requested):
Handling Unit is a doctype, we should treat it like a proper noun. 

"handling unit" => "Handling Unit"

HKuz:
@agritheory all set 👍

PR #63, opened 2023-08-25
Net qty fix

### LINKED ISSUE #52: Netting Handling Unit quantities
Given the following case:

- A Material Transfer for Manufacture Stock Entry requires 50lbs of the Item "Butter".
- We have a Purchase Receipt where we receive 56.26lbs of Butter
- When we scan a the Handling Unit from the Purchase Receipt, the Stock Entry adds a new row for 56.26 lbs, because a row with Butter already exists.

Desired behavior:
If the Handling Unit's stock qty is greater than the required qty, we should assume that it should associate and allow the 50 lbs to be consumed, leaving a net qty associated with that handling unit of 6.26lbs.

If that Handling Unit's stock qty is less than the 50lbs required, it should modify the row to reflect the qty associated with the Handling Unit. We should assume that multiple handling units would be required here. It looks like a [real-world Handling Unit](https://www.bakersauthority.com/collections/bulk-butter/products/cabot-unsalted-butter) for this Item would be a Box of 36 Boxes of (4) 1/4lb sticks. So we should expect to see something like one row that consumes an entire HU and another that consumes 16lbs / 22lb net qty.

Now, here's the problem. When we go to scan the handling unit for that box that we split 16lbs and 22lbs, what is the right answer? The material has been transferred to WIP but it hasn't been consumed. If we want to to cancel the Work Order and return the Items to their warehouse, we may be in a situation where the HU with the net qty has been consumed and we can't "unify" them again.

I think the solution to this is to disallow netting the quantities in Material Transfer for Manufacture but to allow them in Manufacture. Though the case where we have to "unify" a Handling Unit again seems kind of crazy. An alternative would be to create (or prompt) new Handling Units for the items when the Stock Entry has been cancelled. 

Screenshot 2023-07-28 at 16-45-34 Material Transfer for Manufacture - MAT-STE-2023-00001

agritheory:
This fixes ends up being a superset as well as change in strategy from #51, notably that every transfer Stock Entry type creates a new handling unit,. These are notably NOT reversed when the Stock Entry is cancelled. Docs have been update to reflect this.


agritheory:
Can't wait any more to get this one out.

PR #67, opened 2023-09-06
Cancelled SLEs ignored by `get_handling_unit`

### LINKED ISSUE #65: Ignore cancelled Stock Ledger Entries in `get_handling_unit`
In this example we are receiving 40 units of Cocoplum.

| Item             | Warehouse               | Handling Unit  |       Quantity |
| ---------------- | ----------------------- | -------------: | --------------:|
| Cocoplum         | Storeroom               |            123 |          40 Ea |

For the next transactions we will transfer 16 units to a WIP warehouse in a Material Transfer for Manufacture Stock Entry.

| Item             | Warehouse               | Handling Unit  |       Quantity |
| ---------------- | ----------------------- | -------------: | --------------:|
| Cocoplum         | Kitchen                 |            456 |          16 Ea |

The remaining quantity of the original handling unit should show as 24 units.

| Item             | Warehouse               | Handling Unit  |       Quantity |
| ---------------- | ----------------------- | -------------: | --------------:|
| Cocoplum         | Storeroom               |            123 |          24 Ea |

At this point we transfer 6 units of that original handling unit to an arbitrary warehouse with a Material Transfer Stock Entry.

| Item             | Warehouse               | Handling Unit  |       Quantity |
| ---------------- | ----------------------- | -------------: | --------------:|
| Cocoplum         | Other Warehouse         |            789 |           6 Ea |

The remaining quantity of the original handling unit should show as 18 units.

| Item             | Warehouse               | Handling Unit  |       Quantity |
| ---------------- | ----------------------- | -------------: | --------------:|
| Cocoplum         | Storeroom               |            123 |          18 Ea |

If either Stock Entry is Cancelled, there's an issue with how we realistically re-combine the units.

If we cancel the Material Transfer for Manufacture Stock Entry we would expect to see:

| Item             | Warehouse               | Handling Unit  |       Quantity |
| ---------------- | ----------------------- | -------------: | --------------:|
| Cocoplum         | Storeroom               |            456 |         -16 Ea |

| Item             | Warehouse               | Handling Unit  |       Quantity |
| ---------------- | ----------------------- | -------------: | --------------:|
| Cocoplum         | Storeroom               |            123 |          34 Ea |

This is problematic since the qty 16 handling unit is labelled with a ID that doesn't work - scanning "456" will return to results.

The right thing to do I think depends on the transaction, but it boils down to "should we make a new handling unit in this case or not".
- In a scenario where very little time has elapsed since the Stock Entry is submitted, it doesn't make sense to have to re-add the cancelled Handling Units also with a change in Warehouse to make new Handling Units. I imagine this would be very frustrating for someone who just wants to change the Cost Center or has noticed an incorrect Warehouse. 
- In a scenario where the Stock Entry is a Transfer to Subcontractor, more time may pass and the desire to keep the handling units generated by the cancelled transaction is likely stronger.
- Some items may just not allow for physically re-combining successfully (like something that needs to be resealed to be airtight or using our food examples - how would you re-combine a block of cheese?). Giving the user the option to NOT recombine the Handling Unit makes a lot of sense.

agritheory:
- [x] Feature working, design validated
- [x] (corrected net qty) Tests pass
    - [x] There's a new overconsumption failure in another test, not sure if its related or is an artifact from a previous test
- [x] Docs with screenshot of new cancellation workflow


agritheory:
Discussion in #65 should be resolved before merging.

agritheory:
@HKuz Can you take a look at this one? To test cancellation workflow I was setting up a transfer stock entry:
```python
from beam.tests.test_handling_unit import *
test_stock_entry_material_transfer_for_manufacture()
frappe.db.commit()
```
I still need to add to the docs to show the new recombine/ cancellation workflow.
 
@Rgottsch1 I found the bug you were showing, the fix will be included when this is merged.

HKuz:
@agritheory this looks good - two things I noticed, which may be by design but wanted to ask about:
- for items where the entire quantity of the existing HU that are transferred, a new HU is being created (for the test example, the Limequat and Kaduka Lime both get new HUs on transfer)
- after cancel, an item I opted to not recombine kept its HU as expected, but stayed in the Kitchen (wasn't sure if this should go back to the original WH, in this case the Refrigerator)

agritheory:
@HKuz Ready for review

HKuz:
@agritheory docs look good!

HKuz (approved):

PR #88, opened 2024-03-01
Track Subcontracting Process by Handling Unit

viralkansodiya:

PR #92, opened 2024-03-08
Validation: Mandatory handling unit

viralkansodiya:
if stock entry type is not 'Material Receipt'  and Item has enabled handling unit feature then handling unit should be mandatory


viralkansodiya:
Process to merge this PR after merging #88

beam/beam/overrides/stock_entry.py
```diff
@@ -72,3 +72,10 @@ def get_handling_units_for_item_code(doctype, txt, searchfield, start, page_len,
 		.groupby(StockLedgerEntry.handling_unit)
 		.run(as_dict=False)
 	)
+
+
+def validate_items_with_handling_unit(self, method):
+	if self.stock_entry_type != "Material Receipt":
+		for i, row in enumerate(self.items):
```
agritheory: You can use `row.idx` instead and remove `enumerate`.

agritheory (changes requested):

beam/beam/overrides/stock_entry.py
```diff
@@ -72,3 +72,10 @@ def get_handling_units_for_item_code(doctype, txt, searchfield, start, page_len,
 		.groupby(StockLedgerEntry.handling_unit)
 		.run(as_dict=False)
 	)
+
+
+def validate_items_with_handling_unit(self, method):
```
agritheory: To follow Frappe conventions, use 'doc' to refer to the document object when it is not a class method and `self` when you're writing a class method (including overrides).

beam/beam/overrides/stock_entry.py
```diff
@@ -72,3 +72,10 @@ def get_handling_units_for_item_code(doctype, txt, searchfield, start, page_len,
 		.groupby(StockLedgerEntry.handling_unit)
 		.run(as_dict=False)
 	)
+
+
+def validate_items_with_handling_unit(self, method):
+	if self.stock_entry_type != "Material Receipt":
+		for row in self.items:
+			if not row.handling_unit and frappe.db.get_value("Item", row.item_code, "enable_handling_unit"):
+				frappe.throw(f"Row #{row.idx}: Handling Unit is missing for item {row.item_code}")
```
agritheory: Please refactor to f-string and add `frappe._(...)` translation helper

agritheory (changes requested):
Please also rebase

beam/beam/overrides/stock_entry.py
viralkansodiya: Changes has been done as you mentioned.

beam/beam/overrides/stock_entry.py
viralkansodiya: Changes has been done as you mentioned

agritheory (approved):

agritheory:
@viralkansodiya With this feature added, some tests are now failing and we should address that. 

To align with the bakery narrative of the tests, "Water" and "Ice Water" items should not have handling units and we should change this in `tests/setup.py`. 

To fix the repack tests, we only need to add the handling unit key collected [here](https://github.com/agritheory/beam/actions/runs/8627238494/job/23646825549?pr=92#step:11:29) into [this dictionary](https://github.com/agritheory/beam/actions/runs/8627238494/job/23646825549?pr=92#step:11:41). That said, the quality of this test that I originally wrote could be improved by simulating a scan, like we do [here](https://github.com/agritheory/beam/blob/version-14/beam/tests/test_scan.py#L16). 

Let me know if you run into any trouble.

viralkansodiya:
@agritheory 
At this point, we need to modify the validation of mandatory handling units based on the type of stock entry.

For example:

1. If the stock entry type is 'Repack', the handling unit should be mandatory unless it is a finished item.

2. If the stock entry type is 'Manufacturing', the handling unit should be mandatory only for raw materials, not for finished items.

Similarly, we should consider other types of stock entry as well."

Please give your suggestion

Issue #91, opened 2024-03-08
Display Quantity Columns in Recombination Pop-up Window on Cancellation of Stock Entry (Material Transfer Type)


basawaraj-fosserp:
Currently when a Stock Entry (Material Transfer type) is cancelled, a pop-up appears which allows users to select items for which the handling units must be re-combined.

The pop-up window has only three columns: Item Code, Handling Unit, Handling Unit to recombine.

Add additional columns displaying the quantities of each handling units.

The sequence of column should be Item Code, Handling Unit, Quantity, Handling Unit to recombine, Quantity.

![Screenshot 2024-03-08 at 12 00 35 PM](https://github.com/agritheory/beam/assets/103717083/ae6ec027-04d0-4a22-a108-a7171792787b)

viralkansodiya:
@basawaraj-fosserp 
Please review this PR #93

PR #93, opened 2024-03-09
Added remaining qty and transferred qty on dialog box(Cancellation process of stock entry) 

viralkansodiya:
Screenshot 2024-03-09 185754

"As you can see in the dialog box in the above picture, that is the dialog box that pops up during the stock entry cancellation process.

I have added the current quantity of both handling units."

#91


beam/public/js/stock_entry_custom.js
```diff
@@ -100,21 +115,32 @@ async function show_handling_unit_recombine_dialog(frm) {
 	})
```
agritheory, line 115: On line 111, we should change this to 'extra-large'. [Frappe example](https://github.com/frappe/frappe/blob/b331163ce96188f81cf7f5d8a67aa4aab9f92637/frappe/public/js/frappe/utils/diffview.js#L62)

agritheory (changes requested):

beam/public/js/stock_entry_custom.js
viralkansodiya: changes has been done

viralkansodiya:
@agritheory 
any things pending in this PR?

agritheory (approved):

agritheory:
@viralkansodiya Please run the linters again on this branch and push up any changes

PR #104, opened 2024-04-18
fix: add read-only decorator to APIs

### LINKED ISSUE #66: Add @frappe.read_only() to appropriate APIs

Alchez:
Closes #66.

@agritheory I've only added the decorator to whitelisted functions that are not modifying the database. Let me know if I missed anything.


.pre-commit-config.yaml
```diff
@@ -61,8 +61,8 @@ repos:
     hooks:
       - id: validate_customizations
         always_run: true
-        name: .github/validate_customizations.py
-        entry: python .github/validate_customizations.py
+        name: pre-commit/validate_customizations.py
```
agritheory: Don't make this change. It will lead to churn when we update all of the pre-commit scripts later.

agritheory (changes requested):

.pre-commit-config.yaml
Alchez: @agritheory Fair enough, I've reverted that change

Issue #21, opened 2023-06-20
Workstation Integration


agritheory:
From a Workstation, select Work Orders and create/open Job Cards assigned to that Workstation with inventory movement (Transfer for Manufacture, Material Consumption for Manufacture and Manufacture).

hpema:
First Pass design for discussion:

1) Employee gets to the workstation, enters credentials
2) All incomplete task and related work orders show in a list | Also what could be shown is either a finish product from the task or that the task will follow another task, either for same user or the next assigned
3) user selects the job to be completed, the screen loads the items or inventroy needed for the job with its related status. On hand quantity and ready to be used for the job or none. Colour code the lines or ad a badge to show unavailable items.  
4) User scans or selects the inventory thats going to be used and allocate to the workstation. This be the transfer to start the job and confirm stock movement
5) User starts the job by activating the timer, completes the task and moves to the next till completed
6) once all the tasks are completed, job then gets completed
go to step (2)

agritheory:
Specifying the interface a bit, let's assume it's a tablet in landscape orientation. 

> Employee gets to the workstation, enters credentials

I think we want the first pass of this to be simple and we should reuse the built-in login screen. It makes sense that we could support RFID or barcode logins in the future from the same screen.

> All incomplete operations and related work orders show in a list

This can pull from the 'demand' endpoint in its final iteration, but I don't think that's required for now. Let's use the default list view from stonecrop/beam for this. 

> Also what could be shown is either a finish product from the operation or that the operation will follow another operation, either for same user or the next assigned

This would also pull form the "demand" API

> user selects the Work Order to be completed, the screen loads the items or inventory needed for the job with its related status. On hand quantity and ready to be used for the job or none. Color code the lines or ad a badge to show unavailable items.  

I think we can omit or include items that are unavailable at the API level. The thing that makes a MES effective is presentation of instructions (work on this next) and parts availability. I can see how systems will get cluttered with backlogged parts very quickly.

> User scans or selects the inventory that's going to be used and allocate to the workstation. This be the transfer to start the Work Order, Job Card and confirm stock movement

I am imagining an interface here that has a left/right split where the Work Orders are listed on the left and the actions/ inputs /timers are a form-ish view on the right. I'll work with @crabinak to make a mockup.

> User starts the Job Card by activating the timer

At this point I'm wondering if we switch the left-hand Work Order list out with the list of operations.

>completes the Operation and moves to the next till completed

[...]

Issue #15, opened 2023-05-26
Portal/ Mobile interface


agritheory:
### Actions
- Material Transfer
- Manufacture
- Repack
- Send to Subcontractor
- Material Transfer for Manufacture
- Material Consumption for Manufacture
- Receive (Purchase Receipt)
- Ship (Delivery Note)

Components already in progress [here](https://github.com/agritheory/stonecrop)

agritheory:
Splitting some of the workflows here into separate tickets by the user's responsibility.

Shipping and Receiving
 - Purchase Order => Purchase Receipt
 - Sales Return / RMA
 - Send to Subcontractor
 - Pack and Ship

Stocking
- Material Transfer / Material Transfer for Manufacture
- Stock Reconciliation

Manufacturing
 - Material Transfer for Manufacture
 - Manufacture
 - Repack
 - Material Consumption for Manufacture

agritheory:
There are three general patterns for directing the user about what actions need to be taken. These concepts are valuable for building the listview(s) for each core responsibility of BEAM. They can be further filtered by warehouse and/or workstation. 

1. Items have arrived and need to be processed
	1. A package from a supplier has arrived and it needs to be received
	2. A package from a subcontractor(supplier) has arrived and it needs to be received
	3. A package from a customer has arrived and it needs to be processed as a return
2. Items have entered a warehouse that need to be shipped
	1. Typical of manufacturing processes, when done, the items should be placed in a finished goods warehouse to indicate their readiness
	2. In a case where items need to be send to a subcontractor, use the configured workflow for creating shipping documents
3. A newly submitted document has identified items that need to be moved or shipped and are already in the appropriate warehouse
	1. Sales Order has been submitted that indicates that items are ready to be shipped
	2. A Material Request for Manufacture or a Material Request for Transfer has been created (often by a Production Plan) that asks items to be moved from one warehouse to another (usually their default or putaway warehouse to WIP)

<hr>

[...]

PR #106, opened 2024-04-20
Demand - Proof of Concept

agritheory:
This feature aligns in part with cases 2 and 3 of [this comment](https://github.com/agritheory/beam/issues/15#issuecomment-2061985871) of #15. The ideas here need to be refined in order to better understand how it should be used, which is why I'm starting a pull request. As a feature, I think it is likely a prerequisite for performant mobile views, which is why I was developing it in the context of this app, but as a feature, it more strongly aligns with Inventory Tools. We may want to finish the work there and then import it here as a dependency.

The purpose of this feature is to be able to quickly determine if a Work Order or Delivery Note can be started, based on material availability in the required location(s).

I tried several iterations of trying to organize this data in Redis before ultimately giving up - Redis' capabilities for storing structured data is limited and the inability to search by value means that filtering is only possible with the construction a several hash maps, which serve as unsynchronized indexes to other hash maps containing serialized data. It just felt like the wrong tool. 

Refactoring to sqllite meant that I could preserve the tabular nature of the data, add indexes and use an ultimately more familiar approach that is still essentially in-memory. Reads from sqllite are notoriously fast and the write operations should be trivial. We may have to implement a write lock or [ACID files](https://pypi.org/project/acidfile/) approach to make sure the file isn't corrupted in a case where it would be manipulated by more than one user.

<hr>

To be considered feature complete, I think the following has to be added:

- [x] Add nuance to the warehouse(s) signaling demand. This should be configurable and based on a multiselect of warehouse type and parent(s) so as to exclude quality/ quarantine warehouses, subcontractor warehouses and/or third-party warehousing solutions. #107 
- [x] Test suite should integrate this (leveraging `pytest-order`) such that the demand result is re-examined after each stock transaction. This also likely means doing a complete manufacturing cycle in the test suite so that we can track manufactured goods all the way to Delivery Note. #108
- [ ] Create endpoints with filters for `company`, `item_code`, `warehouse`, `workstation` and `assigned` so that this can integrate with #21 
- [x] I've only implemented the bare minimum of updates to the demand table and have not accounted for any logic on how rows should be removed when demand is satisfied. #108
- [ ] This feature should be defeatable in a settings document, probably per-company
- [x] To avoid file corruption concerns, some ACID compliant write wrapper should be added #109 
- [ ] #110 
- [x] #111
- [ ] It would be good to add a "planned availability" feature #157
- [ ] This should be adaptive to Inventory Dimensions, batch and/or serial numbers #168


beam/tests/setup.py
```diff
@@ -1,577 +1,588 @@
...
+		)
+		if i.is_purchase_item and item.get("supplier"):
+			i.append("supplier_items", {"supplier": item.get("supplier")})
+		if i.item_code == "Parchment Paper":
+			i.append("uoms", {"uom": "Box", "conversion_factor": 100})
+			i.purchase_uom = "Box"
+		if i.item_code in ("Water", "Ice Water"):
+			i.append("uoms", {"uom": "Gallon Liquid (US)", "conversion_factor": 15.142})
+			i.purchase_uom = "Gallon Liquid (US)"
+			i.valuation_rate = 0.01 if i.item_code == "Water" else 0.02
```
Alchez: @agritheory there was some conflict in the HU tests, but I wasn't sure which changes to include, so I've added both sets of changes. Can you have a look please?

beam/tests/setup.py
```diff
@@ -1,577 +1,588 @@
...
+	water.stock_entry_type = water.purpose = "Material Receipt"
+	water.append(
+		"items",
+		{
+			"item_code": "Water",
+			"qty": 1000,
+			"t_warehouse": "Refrigerator - APC",
+			"uom": "Gallon Liquid (US)",
+			"basic_rate": 0.15,
+			"expense_account": "5111 - Cost of Goods Sold - APC",
```
Alchez: @agritheory this is included in the changes I mentioned above

beam/tests/setup.py
```diff
@@ -1,577 +1,588 @@
...
+	)
+	water.append(
+		"items",
+		{
+			"item_code": "Ice Water",
+			"qty": 1000,
+			"t_warehouse": "Refrigerator - APC",
+			"basic_rate": 0.30,
+			"expense_account": "5111 - Cost of Goods Sold - APC",
+		},
```
Alchez: @agritheory this is included in the changes I mentioned above

beam/tests/setup.py
```diff
@@ -1,577 +1,588 @@
...
+		)
+		pp.get_mr_items()
+	for item in pp.po_items:
+		item.planned_start_date = settings.day
+	pp.get_sub_assembly_items()
+	start_time = datetime.datetime(settings.day.year, settings.day.month, settings.day.day, 0, 0)
+	for item in pp.sub_assembly_items:
+		item.schedule_date = start_time
+		time = frappe.get_value("BOM Operation", {"parent": item.bom_no}, "sum(time_in_mins) AS time")
+		start_time += datetime.timedelta(minutes=time + 2)
```
Alchez: @agritheory this is included in the changes I mentioned above

beam/beam/demand/demand.py
```diff
@@ -0,0 +1,255 @@
...
+		row.key = frappe.generate_hash()
+		row.delivery_date = str(calendar.timegm(get_datetime(row.delivery_date).timetuple()))
+		row.net_required_qty = str(row.net_required_qty)
+		row.actual_qty = str(get_balance_qty_from_sle(row.item_code, row.warehouse))
+		row.indent = str(0)
+		output.append(row)
+
+		if frappe.db.get_value("Warehouse", row.warehouse, "is_group"):
+			warehouses_to_check = [row.warehouse]
+			warehouses_to_check.extend(get_descendant_warehouses("Ambrosia Pie Company", row.warehouse))
```
agritheory: String literal here

beam/beam/demand/demand.py
```diff
@@ -0,0 +1,255 @@
...
+if TYPE_CHECKING:
+	from erpnext.accounts.doctype.purchase_invoice.purchase_invoice import PurchaseInvoice
+	from erpnext.accounts.doctype.sales_invoice.sales_invoice import SalesInvoice
+	from erpnext.stock.doctype.delivery_note.delivery_note import DeliveryNote
+	from erpnext.stock.doctype.purchase_receipt.purchase_receipt import PurchaseReceipt
+	from erpnext.stock.doctype.stock_entry.stock_entry import StockEntry
+	from erpnext.stock.doctype.stock_reconciliation.stock_reconciliation import StockReconciliation
+
+
+def build_demand_map() -> None:
```
agritheory: Add company argument

agritheory:
@Alchez Can you re-implement the demand action map you made inside of `build_demand_map`? Using the Demand Map report, you can see that transfers and updates aren't implemented. 

The tests are currently passing for me locally, but that's because they're not really testing this enough - the results aren't correct.

Alchez:
> @Alchez Can you re-implement the demand action map you made inside of `build_demand_map`? Using the Demand Map report, you can see that transfers and updates aren't implemented.
> 
> The tests are currently passing for me locally, but that's because they're not really testing this enough - the results aren't correct.

@agritheory, I've created #145 since I wasn't sure about the implementation of the demand maps. We can discuss the map-specific changes there.

agritheory:
Merging features as-is into mobile_v15, but leaving this PR open for completeness

PR #115, opened 2024-05-08
feat: setup demand matrix

### LINKED ISSUE #108: [demand] Create / document decision matrix for updating demanded inventory
Currently there are three documents that can remove inventory demand: Stock Entry, Delivery Note and Sales Invoice (with 'Update Stock' checked). This code needs to be added to the `demand` feature branch and would be best served with a TDD approach since manually executing or evaporating stock transaction is onerous. These documents need to reconcile the demanding Sales Order or Work Order and their satisfaction. There's a subcontracting workflow that also needs to be tested but it will likely be solved with listening on Stock Entries.

The first step for this issue should to design a decision matrix, like the one we've spec'd for [Handling Units](https://github.com/agritheory/beam/blob/version-14/beam/docs/matrix.md). Both of these could benefit from additional clarity about the Stock Entry type, since that does change the behavior.

In prose, I think we should specify it like this:
> When a Stock Entry with a purpose of "Transfer" is submitted, update demand for that Item, subject to warehouse filtering, increasing demand for target warehouse, decreasing demand in source warehouse. 

> When a Stock Entry with a purpose of "Transfer" is cancelled, update demand for that Item, subject to warehouse filtering, decreasing demand for target warehouse, increasing demand in source warehouse. 

With this laid out, the test regime also becomes clear and we can align the desired outcome.

Alchez:
Closes #108


agritheory (approved):
Looks great

PR #121, opened 2024-05-22
Add mobile view for Beam

### LINKED ISSUE #15: Portal/ Mobile interface

Alchez:


Alchez:
@bhattdevarsh let's use components from https://github.com/agritheory/stonecrop/tree/development/beam/src/components into the `Beam.vue` file based on the spec in https://github.com/agritheory/beam/issues/15#issuecomment-2114110285.

Alchez:
@agritheory @bhattdevarsh 

I've added a mirageJS server for mocking, and here's how the UI looks when it's active (currently mirage is always active, but I'll find a better way to toggle that behaviour):

image

bhattdevarsh:
- added `WorkOrder.vue`, `JobCard.vue`, `Operation.vue` pages and its route.
Expected result:

https://github.com/user-attachments/assets/109f432d-0af7-4641-ab29-1407cd9d775f



ToDo: 

- [ ] Operation page is yet to be fixed for mobile view

image

beam/www/beam/pages/JobCard.vue
```diff
@@ -0,0 +1,74 @@
...
+    <h1>Job Cards</h1>
+    <div @click="redirectToJobOperations(operationID = operation)" v-for="operation in operations">
+        <li><span>
+                {{ operation.operationID }}
+            </span>
+            <span class="right-align">
+                {{ operation.completedOperations }}/{{ operation.maxOperations }}
+            </span>
+        </li>
+    </div>
```
Alchez: I think you can use Vue-Router for this instead of defining a function, but test it out:

```suggestion
    <router-link v-for="operation in operations" :to="{ name: 'operation', params: { operationID: operation.operationID } }">
        <li>
        	<span>
                {{ operation.operationID }}
            </span>
            <span class="right-align">
                {{ operation.completedOperations }}/{{ operation.maxOperations }}
            </span>
        </li>
    </router-link>
```

beam/www/beam/pages/JobCard.vue
```diff
@@ -0,0 +1,74 @@
...
+const route = useRoute();
+const router = useRouter();
+const operations = ref([{}]);
+onMounted(() => {
+    // use this to get operations
+    console.log("route.params.orderId ---> ", route.params.orderId);
+
+    // update this value with the correct API call
+    operations.value = [{
+        operationID: "operation 1",
```
Alchez: Can this key just be called `id`? Or do we need to attach the resource name?

beam/www/beam/index.ts
```diff
@@ -0,0 +1,118 @@
+import { createApp } from 'vue'
+import { createMemoryHistory, createRouter } from 'vue-router'
```
agritheory: I don't think the Memory History  API is the right approach.

Alchez: @bhattdevarsh based on #138, let's switch this one to hash mode.

Alchez:
@agritheory @bhattdevarsh I've added API calls to the Work Order, Job Card and Operation pages. Can you check this out on your end?

I think the pending actions are:
- [ ] Start/stop Job Card timer
- [ ] Breadcrumbs? (router back/forward will work in the meanwhile)

Alchez:
@agritheory I've added a `yarn run dev` option that will spawn a page without having to host Frappe, but before we send it to Curt, there's a couple of problems to solve:

- Vite builds the assets into the `www` folder, but the combination of scoped CSS and inherited Frappe stylesheets will make the development setup very different from the production build (as shown below)
- For the API, I was using the `frappe.call` syntax since we were only dealing with fetching resource data, but that breaks in the development setup because the `frappe` namespace won't exist.
  - I can switch to a simple `fetch` request structure to allow mirage to override them, but should I use the resource API (`/api/resource/<doctype>`) or make custom endpoints?

<details>
<summary>Screenshots</summary>

**Development:**
image

**Production:**
image

</details>

agritheory:
@Alchez We can use fetch for now, but this will get refactored to the stonecrop/graphql-client package as we go. 

on the CSS: we want none of the Frappe CSS and should be doing a complete unset and using Stonecrop styles

Alchez:
@agritheory Okay, I've replaced the `frappe.call` instances with the `fetch` API that should work correctly in both Mirage and MySQL contexts.

For the CSS, here's what I've tried:
- Instead of doing a manual unset, I tried creating a new `Website Theme` record that didn't include themes from Frappe and ERPNext. The upside (and downside) of this approach was that **ALL** website pages would never inherit the Frappe themes, including the login page.
- An alternative to apply granular control over the theme on specific pages is to add a script in the `web_include_js` hook that checks for current route and selectively removes the Frappe stylesheet(s).

I've applied the second fix for now, but let me know if you know a simpler approach to unset the stylesheets.

**PS:** Curt doesn't seem to be on this repo, so I'm not able to tag him here or send him a link.

agritheory:
@Alchez I don't think this is an issue. I propose that we include the stylesheet manually in the base component or /beam HTML, not scoped, and then move forward.

agritheory:
Merged demand features in as of aa338fbca059d756a8f96f2cd006dc47d91e52b7, There are still some features to work on but they're out of the critical path

PR #123, opened 2024-05-24
feat: add mirage server to mock workstations

### LINKED ISSUE #112: [workstation] Serialize Workstation data into miragejs for frontend development

Alchez:
Closes #112.

--- 

@agritheory @bhattdevarsh 

I've added a mirageJS server for mocking, and here's how the UI looks when it's active (currently mirage is always active, but I'll find a better way to toggle that behaviour):

image


agritheory:
@Alchez I think we want to check the process and/or frappe.conf.developer_mode to see if the `make_server` call should be included in production or not.

Alchez:
@agritheory Ah that's a good call. I'll add that in.

I also wanted to discuss the UI development piece here, since Frappe's www pages inherit some styling that seems to be have been used. I'm not sure how we want to harmonize the built Frappe page with the development workflow. We can get on a call to discuss that.

agritheory:
@Alchez We should include an unset utility, [like this](https://nikolasbarwicki.com/articles/remove-all-styling-with-one-line-of-css/)

PR #132, opened 2024-06-14
Descendant warehouse lookup with warehouse types

### LINKED ISSUE #107: [Demand] Warehouse descendant lookup
From #106: 
> Add nuance to the warehouse(s) signaling demand. This should be configurable and based on a multiselect of warehouse type and parent(s) so as to exclude quality/ quarantine warehouses, subcontractor warehouses and/or third-party warehousing solutions.

<br>

ERPNext has an imperfect implementation of this idea in the Auto Reorder abstraction. Its shortcoming include that it's not company sensitive (which can be an advantage, I suppose). One of the features that _is_ implemented is its ability to check in descendant warehouses. That said, not all warehouse types should be checked and I think it makes sense to require a configuration that one or more warehouses have a specific Warehouse Type(s) that allows the user to filter for the correct goods.

MyuddinKhatri:
closes #107 

- Descendant warehouse lookup with warehouse types
- Added Warehouse Types Multiselect in Beam Settings 
image

- While looking for descendant warehouses this will exclude the warehouses with the warehouse types mentioned in the BEAM Settings

### DIFF
--- a/beam/beam/demand/demand.py
+++ b/beam/beam/demand/demand.py
@@ -185,3 +185,39 @@ def modify_demand(
 					)
 
 		conn.commit()
+
+
+def get_descendants_of_warehouse_with_warehouse_types(beam_settings, warehouse):
+	beam_settings = frappe.get_doc("BEAM Settings", beam_settings).as_dict()
+	warehouse_types = []
+	if beam_settings.warehouse_types:
+		warehouse_types = [wt.warehouse_type for wt in beam_settings.warehouse_types]
+
+	if warehouse_types:
+		order_by = "lft"
+		limit = None
+		lft, rgt = frappe.db.get_value("Warehouse", warehouse, ["lft", "rgt"])
+
+		if rgt - lft <= 1:
+			return []
+
+		warehouses = frappe.get_list(
+			"Warehouse",
+			{
+				"lft": [">", lft],
+				"rgt": ["<", rgt],
+				"company": beam_settings.company,
+				"warehouse_type": ["not in", warehouse_types],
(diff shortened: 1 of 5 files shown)


agritheory:
@MyuddinKhatri The description language isn't clear to me - does this exclude the selected warehouse types or include only the selected warehouse types? 

If the logic is exclude - "These Warehouse Types will be excluded when checking for inventory availability"

MyuddinKhatri:
> @MyuddinKhatri The description language isn't clear to me - does this exclude the selected warehouse types or include only the selected warehouse types?
> 
> If the logic is exclude - "These Warehouse Types will be excluded when checking for inventory availability"

@agritheory As it excludes I'll change it to the text you've mentioned above.

MyuddinKhatri:
@agritheory As discussed will implement - show quantities in the descendant warehouses with indentation in the report.

MyuddinKhatri:
@agritheory This is how it looks, 

image
right now it is showing flat tree (parent -> all nested children in one indent), do we want to show exact tree structure for nested children?
One more thing, what do you think about data for Net req qty in child warehouses?

agritheory:
> do we want to show exact tree structure for nested children? 

I think showing the demand:location relationship is enough. 

> What do you think about data for Net req qty in child warehouses?

We should show required quantity and UOM from specific WO/SO and total demand from all WO/SO. This report isn't likely to be the primary interface for this feature, it will be more of a debugging and analysis tool.

MyuddinKhatri:
Added Stock UOM column in the report
image

PR #137, opened 2024-07-15
135 preserve complete hu

robproject:
### Description
Adds setting to allow preservation of complete Handling Units for those transactions where they are complete quantities.
(mat xfer, Send to Subcontractor, ~~mat xfer for mfg~~)

Existing behavior with generation for every transaction is default (unchecked) for this setting.

**Modifies get_handling_unit method** since Handling Units are no longer implicitly grouped by Warehouse due to being shared via full qty transfers, the query was rewritten to include the Warehouse groupby. The newest created SLE is returned (this needs to be ordered by posting date as well) for a warehouse with a qty, since there will only ever be one warehouse with any qty for a HU.

### Testing
Observe HU is preserved between source and target in the above cases when 'Always Generate for Transfer' is unchecked in BEAM Settings.

~~note: pytest was not created~~

image


agritheory:
@robproject This will require tests, let me know if you want some suggestions about it.

robproject:
Yes, I was thinking verifying HU is preserved or not based on those three SE types and the setting would satisfy the requirement, should there be others?

agritheory:
In a case where anything less than the full amount is transferred you end up with new HUs, which I think makes the most sense. I would want this to be under test as well.

robproject:
if the HUs have any change in qty, would it be a repack or non-transfer SE type?

agritheory:
Any type of transfer should follow this rule.

robproject:
Since the preservation is confined to those three SE types, it can have an additional condition added for preservation that verifies the HU is the complete qty. What if it 1. fetched the SLE for adding that qty comparison to the condition for preservation, or 2. fetched it to compare qty and throw an error in validate_items_with_handling_unit ?

beam/beam/handling_unit.py
```diff
@@ -42,9 +42,12 @@ def generate_handling_units(doc, method=None):
...
 			and row.handling_unit
 		):
-			handling_unit = frappe.new_doc("Handling Unit")
-			handling_unit.save()
-			row.to_handling_unit = handling_unit.name
+			if settings.always_generate_for_transfer:
+				handling_unit = frappe.new_doc("Handling Unit")
+				handling_unit.save()
+				row.to_handling_unit = handling_unit.name
+			else:
```
agritheory: This should include a condition where the stock quantities are equal. If you are splitting a HU into two pieces, they cannot have the same number.

agritheory:
0. Transfer of 20Eac from HU of 20Ea to another warehouse - re-use the HU. I believe this should apply to "Material Transfer", "Send to Subcontractor" and "Subcontracting Receipt". "Material Consumption for Manufacture" / process manufacturing generally is not under test in BEAM.  

1. Transfer of 15Eac from HU of 20Ea to another warehouse - new HU; Transfer of remaining 5Eac from HU of 20Ea to another warehouse - new HU. I believe this should apply to "Material Transfer", "Send to Subcontractor" and "Subcontracting Receipt". "Material Consumption for Manufacture" / process manufacturing is not under test in BEAM.

2. Manufacture SE with only one input item (like painting) should be eligible to re-use, but this should be configured at the BOM, since it won't always be correct. This seems like a special case and could be considered out of scope. We don't have a BOM or story for this case, but could create one by having a "formed" pie crust Item that only requires baking.

3. Repack cannot reuse the HU since it is by definition split into new HUs or it is literally repacked, so the intrinsic definition of a HU has changed. 

4. Packing Slip, Delivery Notes and Sales Invoices can only consume HU and if they consume all of it, its gone. Partial consumption will allow for net quantities.

robproject:
```
 1. Transfer of 15Eac from HU of 20Ea to another warehouse - new HU; Transfer of remaining 5Eac from HU of 20Ea to 
 another warehouse - new HU. I believe this should apply to "Material Transfer", "Send to Subcontractor" and
 "Subcontracting Receipt". "Material Consumption for Manufacture" / process manufacturing is not under test in BEAM.
```
Would the second action of xfer remaining 5Eac be equal to an xfer of that HU? Scanning after the xfer of 15Eac would mean that HU would populate as 5Eac?

agritheory:
> ```
>  1. Transfer of 15Eac from HU of 20Ea to another warehouse - new HU; Transfer of remaining 5Eac from HU of 20Ea to 
>  another warehouse - new HU. I believe this should apply to "Material Transfer", "Send to Subcontractor" and
>  "Subcontracting Receipt". "Material Consumption for Manufacture" / process manufacturing is not under test in BEAM.
> ```
> 
> Would the second action of xfer remaining qty 5 be equal to an xfr of that HU? 

Potentially, yes. The "and its packaging " part of the definition of Handling Unit complicates this. A "Sack of Flour" item - sure, the sack and remaining flour are consumed. If it's a six-pack of Corona Light, I think there's an argument that you discard the paper holder at some point. This might be an argument for configuring "always split" at the item level as well as a global fallback. 

> Scanning after the xfer of qty 15 would mean that HU would populate as qty 5?

That's how net quantities should behave, yes. But this currently doesn't preserve the HU of those final five items into the new location. I am inclined to think that it should not.

robproject:
Inventory dimension preservation, including HU, can be configured in a separate app

PR #136, opened 2024-07-01
ZPL Utils

agritheory:
This PR provides a couple of ZPL related utilities and POC for BEAM.

<details><summary>A print preview using a jinja method and the free Labelary API</summary>

image

</details>

<details><summary>A POC for click-to-print</summary>

This utilizes a CUPS instance that is on the same network or localhost to the ERPNext instance. Authentication if required would be a stretch goal. 

image

</details>

<details><summary>A ZPL and jinja mixed environment</summary>

This allows the use of the helpers from the [py-zebra-zpl](https://github.com/mtking2/py-zebra-zpl) library for object-orient label composition.

image

```
{% set hu = get_handling_unit(doc.name) %}
{% set label = zebra_zpl_label(width=6*203, length=4*203, dpi=203, orientation="R") -%}

{{ label.add(zebra_zpl_barcode(hu.handling_unit, width=4, height=260, position=(120, 40), justification="C", ratio=1, human_readable='N', orientation="N")) -}}
{{ label.add(zebra_zpl_text(hu.handling_unit, position=(40, 320), width=(4*203-80), font_size=40, justification="R", orientation="N")) }}
{{ label.add(zebra_zpl_text(frappe.utils.cstr(hu.qty) + " " + hu.uom, position=(120, 400), width=(4*203-80), font_size=40, orientation="N")) }}
{{ label.add(zebra_zpl_text(hu.item_code, position=(120, 480), width=(4*203-80), font_size=40, orientation="N")) }}
{{ label.add(zebra_zpl_text(hu.warehouse, position=(120, 560), width=(4*203-80), font_size=40, orientation="N")) }}
{{ label.add(zebra_zpl_text(frappe.utils.format_datetime(hu.posting_datetime), position=(120, 640), width=(4*203-80), font_size=40, orientation="N")) }}

{{ label.dump_contents() }}
```

</details>

<details><summary>It almost works, I am having an issue getting the orientation correct. </summary>

image

</details>

PR #140, opened 2024-07-19
docs: add documentation for ZPL utils

### LINKED ISSUE #139: Add documentation for ZPL utils features

Alchez:
Closes #139.

### DIFF
--- a/beam/docs/zebra_printing.md
+++ b/beam/docs/zebra_printing.md
@@ -1,7 +1,7 @@
 # Zebra Printing
 
 To create a Zebra print format, you need the following documents:
-- A ZPL Print Format made against Handling Units that uses the available Jinja utility functions to generate ZPL code.
+- A ZPL Print Format made against Doctype that may contain barcodes (Item, Warehouse, Handling Units, etc.) that uses the available Jinja utility functions to generate ZPL code.
 - A document Print Format that uses the free Labelary API to convert the above ZPL code and generate a preview of the print output for the linked document.
 
 ### ZPL Code Generation
@@ -12,6 +12,8 @@ Currently, only three types of printable ZPL data can be generated with utilties
 - `Label`
 
 Beam uses the [py-zebra-zpl](https://github.com/mtking2/py-zebra-zpl) library to generate the above types, as it provides a basic interface to create ZPL code using Python objects. Please refer to the library's documentation for more information on how to use it.
+
+**Note:** Additional ZPL elements (like graphic fields) and commands (text mirroring, character encoding, etc.) can be developed separately and added as text directly to the ZPL Print Format. For more information, visit the [official documentation page](https://supportcommunity.zebra.com/s/article/ZPL-Command-Information-and-DetailsV2?language=en_US) or the [Labelary ZPL Programming Guide](https://labelary.com/zpl.html).
 
 In addition, Beam exposes the following Jinja functions to be used within a Print Format:
 
@@ -25,7 +27,7 @@ Generate a [Code 128](https://en.wikipedia.org/wiki/Code_128) barcode image. It 
 
 ##### Example
 ```jinja
-{{ barcode128('123456') }}
+{{ barcode128(doc.barcodes[0].barcode) }}
 ```
 
 ---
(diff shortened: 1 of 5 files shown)


beam/docs/zebra_printing.md
```diff
@@ -0,0 +1,178 @@
+# Zebra Printing
+
+To create a Zebra print format, you need the following documents:
+- A ZPL Print Format made against Handling Units that uses the available Jinja utility functions to generate ZPL code.
```
agritheory: It doesn't need to be a Handling Unit, but that's the example we include. It would be very reasonable to have a ZPL format for another barcoded doctype (Item, Warehouse, etc)

beam/docs/zebra_printing.md
```diff
@@ -0,0 +1,178 @@
...
+- A document Print Format that uses the free Labelary API to convert the above ZPL code and generate a preview of the print output for the linked document.
+
+### ZPL Code Generation
+
+Currently, only three types of printable ZPL data can be generated with utilties within Beam:
```
agritheory, line 9: What is the caveat about Graphics fields? Perhaps we should recommend they be developed separately and included as plain text?

beam/docs/zebra_printing.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### `barcode128`
+
+Generate a [Code 128](https://en.wikipedia.org/wiki/Code_128) barcode image. It takes the following arguments:
+
+- `barcode_text`: The text to be encoded in the barcode. Required.
+
+##### Example
+```jinja
+{{ barcode128('123456') }}
```
agritheory: Can we change the literal here to something more instructive, like `doc.barcodes[0].barcode`

beam/docs/zebra_printing.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### `formatted_zpl_barcode`
+
+Generate a formatted ZPL barcode. It takes the following arguments:
+
+- `barcode_text`: The text to be encoded in the barcode. Required.
+
+##### Example
+```jinja
+{{ formatted_zpl_barcode('123456') }}
```
agritheory: Let's remove the string literal here as well

beam/docs/zebra_printing.md
```diff
@@ -0,0 +1,178 @@
...
+- `dpi`: The DPI of the printer. Defaults to 203.
+  - Visit the [official documentation](https://supportcommunity.zebra.com/s/article/000026166) to determine the DPI for your Zebra printer model.
+- `print_speed`: The print speed of the printer in inches per second (ips). Defaults to 2.
+  - Slower print speeds typically yield better print quality.
+  - Visit the [official documentation](https://supportcommunity.zebra.com/s/article/Setting-the-Print-Speed-via-ZPL) to determine the acceptable print speed values for your Zebra printer model.
+- `copies`: The number of copies to print. Defaults to 1.
+
+##### Example
+```jinja
+{% set label = formatted_zpl_label(width=6*203, height=4*203, dpi=203) %}
```
agritheory, line 65: Is it worth mentioning that the math here is inches * DPI?

beam/docs/zebra_printing.md
```diff
@@ -0,0 +1,178 @@
...
+{{ label.start }}
+{{ label.end }}
```
agritheory, line 68: A comment about ZPL Code here, except that raw printing doesn't have a way to add comments

agritheory (changes requested):
A couple of small changes. I think we need to include instructions on setting up the Zebra printer in CUPS: https://supportcommunity.zebra.com/s/article/Adding-a-Zebra-Printer-in-a-CUPS-Printing-System?language=en_US

beam/docs/zebra_printing.md
Alchez: Oh, I had restricted it to Handling Unit since that's the filter applied in the click-to-print POC you added in #136. I'll change the wording to something more generic.

beam/docs/zebra_printing.md
Alchez: I think so. I've added it to the docs for this function directly.

beam/docs/zebra_printing.md
Alchez: I've added a line about arbitrary ZPL code, but I think it should be illustrative enough that the user can decide how to use it.

PR #145, opened 2024-08-02
fix: apply demand map to transactions

Alchez:
Based on https://github.com/agritheory/beam/pull/106#issuecomment-2264292786.

I'm creating a separate PR because I'm not sure about the implementation yet and this needs some discussion.


agritheory:
@Alchez When I reinstall, there's an error here where demand effect is undefined. `build_demand_map` and `build_allocation_map` need to be applicable to both a document context and to a "build from scratch" context. 

```
  File "/home/tyler/pinyon/apps/frappe/frappe/commands/utils.py", line 269, in execute
    ret = frappe.get_attr(method)(*args, **kwargs)
  File "/home/tyler/pinyon/apps/beam/beam/tests/setup.py", line 46, in before_test
    build_demand_map()
  File "/home/tyler/pinyon/apps/beam/beam/beam/demand/demand.py", line 143, in build_demand_map
    build_allocation_map()
  File "/home/tyler/pinyon/apps/beam/beam/beam/demand/demand.py", line 251, in build_allocation_map
    if demand_effect == "increase":
UnboundLocalError: local variable 'demand_effect' referenced before assignment
```

beam/beam/demand/demand.py
```diff
@@ -134,64 +167,206 @@ def build_demand_map() -> None:
...
+	action: dict | None = None,
+):
+	item_query = warehouse_query = ""
+	if row:
+		item_query = f"WHERE item_code = '{row.item_code}'"
+		# TODO: should we only consider warehouse if it is the same as the demand warehouse?
+		# if action:
+		# 	warehouse_field = action.get("warehouse_field")
+		# 	if warehouse_field:
+		# 		warehouse_query = f"AND warehouse = '{row.get(warehouse_field)}'"
```
Alchez: @agritheory this part of the code was checking the transaction's warehouse against the demand warehouse, which I'm not sure if that's intentional? I've commented it out and that makes the allocations against non-demand warehouses to work correctly.

agritheory: Yeah, part of the demand feature is that it is only satisfied when it's in the correct warehouse. We can allocate before that, but the demand cannot be met unless the inventory is in the correct warehouse.

Alchez: So should I keep that code? Keeping it will only add inventory allocation in demand warehouses.

agritheory: No, allocations are warehouse-agnostic except for the excluded warehouses (quarantine /scrap/ subcontractor). We want to allocate from all warehouses except those. So I think that this code needs to be updated for that scenario but not a 1:1 matching one.

Alchez: Sounds good. I'll make some changes tomorrow.

Alchez:
@agritheory Okay, this is ready for review. Can you give it a test as well?

PR #146, opened 2024-08-02
fix: create handling units after submit

### LINKED ISSUE #144: Generate Handling Unit should be happening in `on_submit` hook instead of `before_submit` hook
But this causes the tests to fail in way that where it isn't obvious why - it appears that handling units aren't being generated

Alchez:
Closes #144 

---

I've changed the logic for setting the Handling Units directly into the database after creation.

image

### DIFF
--- a/beam/beam/barcodes.py
+++ b/beam/beam/barcodes.py
@@ -6,7 +6,7 @@ import barcode
 import frappe
 from erpnext import get_default_company
 
-from beam.beam.doctype.beam_settings.beam_settings import create_beam_settings
+from beam.beam.doctype.beam_settings.beam_settings import get_beam_settings
 
 
 @frappe.whitelist()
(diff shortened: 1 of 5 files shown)


Alchez:
@agritheory So based on the failing tests, it seems like the Handling Units were set to be created on `before_submit` because they need to pass to the SLEs, which are created after submit.

If the HUs are created after submit as well, they won't be available in the SLEs anymore. I'm not sure what a good fix here is.

What do you think?

agritheory:
I think we're back to `before_submit`. We just want to make that `before_submit` logic happens after the db rollback setpoint if there's an error in `submit`.

Alchez:
@agritheory

> We just want to make that before_submit logic happens after the db rollback setpoint if there's an error in submit.

That makes sense. Let me check if that's already happening or if a change is required.

Alchez:
@agritheory It seems like the rollback is happening correctly if a failure happens in `before_submit` or `on_submit`. Is there a case you've seen where this behaviour isn't happening?

PR #148, opened 2024-08-05
docs: add network printer settings example

### LINKED ISSUE #141: Configure default and secondary Network Printer Settings for Stock transaction doctypes and Workstations
Let's create an example Network Printer Settings in the test suite for demo data, bypassing whatever validations are needed, because we won't be configuring a Zebra printer in CUPS in the CI. 
No tests required for this workflow, screenshots will be enough.

HKuz:
Closes #141 for `version-14`


beam/tests/setup.py
```diff
@@ -570,3 +571,17 @@ def create_purchase_receipt_for_received_qty_test(settings):
 		},
 	)
 	pr.save()
+
+
+def create_network_printer_settings(settings):
+	printer_settings = [
+		{"name": "Primary Example Printer", "server_ip": "localhost", "port": 8888},
```
agritheory: Receiving Printer and Kitchen Printer

beam/tests/setup.py
```diff
@@ -570,3 +571,17 @@ def create_purchase_receipt_for_received_qty_test(settings):
...
 	)
 	pr.save()
+
+
+def create_network_printer_settings(settings):
+	printer_settings = [
+		{"name": "Primary Example Printer", "server_ip": "localhost", "port": 8888},
+		{"name": "Secondary Example Printer", "server_ip": "localhost", "port": 9999},
+	]
+	for ps in printer_settings:
```
agritheory, line 581: Check if these records already exist `if frappe.db.exists`

beam/tests/setup.py
HKuz: @agritheory - updated in latest push (both versions are good)

PR #149, opened 2024-08-09
feat: requires login for /beam

### LINKED ISSUE #142: [mobile v15] Require login for the /beam path
We will want to wrap the Frappe login process here so we can provide alternative login hooks, like the ability to scan in when there's (for example - not in score) a matching barcode and with a matching IP address or a matching custom header.

MyuddinKhatri:
closes #142


beam/www/beam/index.ts
```diff
@@ -94,6 +107,19 @@ const router = createRouter({
 	routes,
 })
 
+router.beforeEach((to, from, next) => {
```
agritheory, line 115: I think we want to detect if the Vite server is running and opt out of this for development. This may not be a great idea for security purposes, so lets try to pen test it with some console hacking. May need @Alchez work on #121

agritheory (commented):
See comment

beam/www/beam/index.ts
Alchez: @agritheory @MyuddinKhatri I think this _may_ be fine this way since the entrypoint for the development setup (`/dev.html`) is separate from the production setup (`/beam`) anyway.

But if there's a part of the development setup that would need some production setup, then we can look at the Vite metadata. Vite [recommends](https://vitejs.dev/guide/env-and-mode) checking against the `import.meta.env.DEV` and `import.meta.env.PROD` keys for development vs production setups.

Alchez (approved):

Alchez:
@agritheory @MyuddinKhatri To move this one ahead, I've set the relative hash paths such that they would work identically for both dev and prod environments.

PR #152, opened 2024-08-14
fix: eval basic_rate

robproject:
### Description
Typo assigns 'basic_rate' as value for 'field' property instead of comparison.

### Testing
Observe 'add_or_associate' can be extended and target values are updated.


agritheory (approved):
Nice catch

PR #162, opened 2024-08-20
fix: update demand when sales or work orders are updated

### LINKED ISSUE #143: [demand] Add demand when Work Order or Sales Order is submitted
Demand (not allocation) should be added when a new Work Order or Sales Order is submitted. It should be removed when the Work Order is completed or manually closed, or the Sales Order is satisfied via Delivery Note or Sales Invoice or manually closed.

Alchez:
Closes #143 

---

Demand (not allocation) should be added when:
- [x] a new Work Order is submitted 
- [x] a new Sales Order is submitted

Demand (not allocation) should be removed when:
- [x] a Work Order is completed
- [x] a Work Order is manually closed
- [x] a Sales Order is satisfied via Delivery Note
- [x] a Sales Order is satisfied via Sales Invoice
- [x] a Sales Order is manually closed

Todo:
- [x] Timestamp priority: delivery_date / expected_delivery_date, creation, name


Alchez:
@agritheory Currently the approach I've used is very brute-force, in which I completely rebuild the demand and allocation tables. I think, but I'm not sure, that this may be destructive in case user actions later change allocations manually.

If that is true, should this instead be incremental changes? And should we discourage ever truncating the tables in any operation?

agritheory:
@Alchez Yes, we want this to be incremental and in the future to allow for manual allocation.

Alchez:
@agritheory This one is ready for a review, and I'll need your help with the tests.

agritheory:
> @agritheory This one is ready for a review, and I'll need your help with the tests.

Sure, can you describe how you were testing it manually? These tests are tricky because we want to test for changes in the demand db in parallel to testing for correct net quantities for handling units, so it does end up being adding to existing tests rather than writing new ones.

Alchez:
@agritheory I was using the following cases to test, but it's probably not an exhaustive list:
- Submit Sales Order/Work Order (expected = actual = demand increases)
- Cancel Sales Order/Work Order (expected = actual = demand decreases)
- Submit Delivery Note (expected = actual = demand decreases + stock allocated)
- Cancel Delivery Note (expected = actual = demand re-created + allocated stock removed)

agritheory:
> Submit Sales Order/Work Order (expected = actual = demand increases)

I think this is covered in the existing demand tests. 

> Cancel Sales Order/Work Order (expected = actual = demand decreases)
> Cancel Delivery Note (expected = actual = demand re-created + allocated stock removed)

This needs a new test that should instantiate a Sales Order and then cancel it. Copying the one that is there is fine. Asserts should be on the increased qty and then the reduced qty after cancelling. Adding the satisfying inventory via "Material Receipt" and then testing the DN submit / cancel as part of the same test or a subsequently ordered test would be good.

> Submit Delivery Note (expected = actual = demand decreases + stock allocated)

I think this can use the existing test and assert on the demand side effect here: https://github.com/agritheory/beam/blob/demand/beam/tests/test_handling_unit.py#L324

Alchez:
@agritheory I've updated and added some tests (that atleast seem to be passing in the CI), but I'm not sure about the output of the DN-cancel test. Let's discuss the expected values there today.

agritheory:
Root cause on the determinism issue we were seeing came from the production plan returning the items from the Sales Order sorted by name, which was regenerated every time. I added a sort to this after that and then every thing worked.

PR #167, opened 2024-09-06
feat: implement scan matrix for mobile

### LINKED ISSUE #158: [mobile] Refactor scanning decision matrix to apply to list views in Mobile context
We want to apply this programmatically so that overrides can be handled appropriately.

Alchez:
Closes #158. Requires https://github.com/agritheory/stonecrop/pull/149.


Alchez:
@agritheory I've added a `scan.ts` composable that can be manually used within each component. Currently it sets up the list-form context based on the current route's doctype (which I've defined in the route config).

There's a few things to work out still:
- [x] **Form data:** Frappe uses `cur_frm.doc` to expose form data directly into the browser. Since we're manually using the composable, I'm wondering if we should just fetch and pass the form data to the composable to use.
- [x] **Form actions:** The base `cur_frm` namespace also performs actions like adding/removing child table rows, refreshing fields, etc. We might want to re-think that in the mobile context since we're using Vue.

beam/www/beam/routes.ts
```diff
@@ -0,0 +1,80 @@
...
+		path: '/work_order/:orderId/operation/:id',
+		name: 'operation',
+		component: Operation,
+		meta: { doctype: 'Operation', view: 'form', requiresAuth: true },
+	},
+	{
+		path: '/receive',
+		name: 'receive',
+		component: Receive,
+		meta: { doctype: 'Purchase Order', view: 'list', requiresAuth: true },
```
agritheory: 'Purchase Receipt'

beam/www/beam/scan.ts
```diff
@@ -0,0 +1,243 @@
...
+					return (
+						(row.item_code === field.context.item_code && row.stock_qty === field.context.stock_qty) ||
+						row.handling_unit === field.context.handling_unit
+					)
+				})
+			) {
+				if (!doc.items.length || !doc.items[0].item_code) {
+					doc.items = []
+				}
+				let child = cur_frm.add_child('items', field.context)
```
agritheory: We need to reimplement a helper for this that uses Pinia

agritheory (changes requested):
@Alchez I think you've gotten a head start on the reviewed changes before I submitted this. Let me know when to re-review.

Alchez:
@agritheory There's a couple of things left to do here:
- Make sure that adding a new row mirrors the API calls made in Frappe's forms
- Add an ability to save changes on the client and have it update the database via the API

And maybe more in case I've missed anything. But you can give it a fresh review until I can get to those.

Alchez:
@agritheory I made another pass at this, and I think I'll need to brainstorm a couple of things here:

- **Adding new rows:**
  - Whenever a scan outputs a new row in the table, we need to figure out a way to bootstrap an empty row and then add details on top of it.
  - Frappe uses the `frm.add_child` and `doc.append()` APIS in the client and server respectively to build the empty rows, and then update any user inputs onto it.
  - Should we consider adding default empty row templates (for ALL child tables?) in the initial server context that'll let us easily create rows?
- **Component layout:**
  - I'm trying to figure out a proper layout for the components in the application layer, and I've created a diagram that should help visualise how the different elements would work. Does it make sense for you?

image

agritheory:
> Whenever a scan outputs a new row in the table, we need to figure out a way to bootstrap an empty row and then add details on top of it.

I think having a helper function that add a new object with the keys is the thing to do here. We know the keys from the Type. Passing the entire array to the server to add a row is not the correct solution.

> Diagram

In the root component should be navigation, scan input and the router view. The Pinia store is instantiated here and only here; it is accessed via composable in the nested views like WorkOrder, Transfer and Ship

Alchez:
@agritheory Okay, I've setup the scan handler class and the Pinia store to be able to work across views. I think the only thing remaining task here is to add a 'Save' element that will allow passing the changes to the database. How do you think that should be handled?

Also, for the "add new row" problem, it seems like most keys aren't required when you're adding a child table row. So I was able to get away with simply adding the item/barcode-related keys and letting Frappe handle the rest. I'm sure this logic will need to be revisited later, but it should be good enough for the doctypes we're using right now.

agritheory (approved):
Let's get this merged so we can incorporate state management in other contexts as well. Looks great.

PR #169, opened 2024-09-07
[mobile] fix work order routing

agritheory:


agritheory:
@Alchez How do we want to add typing to this change? Better to vendor it?

@lauty95 FYI

Alchez:
@agritheory I think vendoring it might be more practical since Frappe doesn't expose types and we can cherry-pick the objects we want typed too. I've noted it over at https://github.com/agritheory/beam/pull/170#pullrequestreview-2288869044 but I think this might also need to mocked for the dev setup.

PR #170, opened 2024-09-07
[mobile] Refactor WO view

agritheory:
WIP: 
image

Issues: 
Styling of navbar / primary action (stonecrop #)
Styling of buttons
Not using new metadata component
Operations and Items do not stack in portrait mode

image


Alchez (approved):
@agritheory The changes generally look good, but I noticed a few minor things that we may want to consider?

- The items list is setup with `ListAnchor`, which will link to `/` if a route is not provided. So clicking on any of the items just takes you to the home page.
- You've used `frappe.scrub` in this PR, but I think that may cause an issue with the dev setup. But is that something we should look at right now even?

agritheory:
@Alchez Good catch. I'll fix these in my next pass through.

agritheory:
To dos after iterating on this a little:
- When a ListView is empty we should show a message or provide slot for a component. This is particularly applicable for the new Move page (blank Stock Entry) and what we want to do with Repack.
- The Transfer component's buttons should be moved to their own component @lauty95. We need some styling on these too, I will ask @crabinak for some help. 
- We're at the point where we need to start incorporating state management throughout and try to use that as primary interface to the server
- Haven't tested scanning, but we may need to add some fixturing to barcodes for consistency in tests 
- I managed to break a new thing while re-baseing but nothing looks wrong with the code @Alchez

Alchez:
@agritheory I was able to fix up the bugs, and I made some additional changes as well:
- Expanded the data store to include document workflow actions like Save, Submit, Cancel, etc.
- Simplified `Transfer.vue` to use the store instead of component methods
- Additional typing and tooling improvements
- Updated dependencies to latest (to use Vue 3.5)

Issue #165, opened 2024-08-29
[mobile] Add hooks for custom routes and components in BEAM


agritheory:
These can serve custom components, but we'll need to figure out how to do that. Probably by building in the customer app to `window.beam.components['componentName']`. The extra routes can be handled at `window.beam.routes` (an array).

This should almost certainly happen in the build step, which will mean some kind of parsing of hooks.py with a node (pre) build script to include the extra files. I'd rather go through this effort than introduce a separate build script or a customer's build script, but that may be required.

PR #171, opened 2024-09-08
Component and Route cross-application import

agritheory:
This is a POC for automatically importing custom components into the BEAM app. #165 

TBD: the actual structure of the hook in `customer_app/hooks.py`

This PR does not address any custom routing requirements, which is very likely required.


Alchez:
@agritheory I think we have a few additional scenarios here (that may be inclusive with each other):

- **Component discovery:** If a user wants to expose a directory under which their custom Beam components would live, then we should be able to apply that via hooks too.
  - I think the structure of the hooks might be an array of paths to match the `dirs` property of [`unplugin`](https://github.com/unplugin/unplugin-vue-components#configuration).
  - This would allow users to develop new components in a single place without having to manually register each component, and also allow over-writing existing Beam components.
  - We could also allow users to exclude components from being registered (either explicitly via the config or by disabling sub-directory discovery and nesting components in a different or sub-directory).
- **Routing:** You've already covered this in #165 but additional routes could also be over-written or extended on top of the default Beam routes.

---

I'm thinking the hooks structure could look something like the following (if we wanted pure Python):
```python
# hooks.py
beam_override = {
	"components": ["./custom_app/beam/components/"],
	# (or alternatively)
	# "components": {"Repack": "./custom_app/beam/components/Repack.vue"},
	"routes": [
		{
			path: '/repack',
			name: 'repack',
			component: "Repack.vue",
			meta: { "doctype": "Stock Entry", "requiresAuth": True },
		}
	],
}
```

or something like this (if we wanted to process JS/TS files directly instead):
```python
# hooks.py
beam_override_file = "public/js/beam/setup.ts"
```

```typescript
// setup.ts

// if we wanted to supply a directory
const dirs = ["./custom_app/beam/components/"]

// if we wanted to explicitly define components
import Repack from "./custom_app/beam/components/Repack.vue"
const components = { Repack: Repack }

// if we wanted to over-write/extend routes
const routes = [
	{
		path: '/repack',
		name: 'repack',
		component: Repack,
		meta: { "doctype": "Stock Entry", "requiresAuth": true },
	}
]

export {
	dirs,
	components,
	routes
}
```

The latter approach might also allow us to pass existing context to do stuff like [dynamic routing](https://router.vuejs.org/guide/advanced/dynamic-routing.html).

This hook would be read in the Vite entry-file and the routes and components set up based on the priority order of: Custom App > Beam (default).

What do you think?

agritheory:
@Alchez `hooks.py` Is the appropriately default configuration file for a Frappe project, I think it's where we want to be. If functions are required, they can be serialized to JSON, which we're already doing in ATable's `format` function for example. 

I it does make sense to respect a "just a directory" approach as well as listing individual files. These approaches can be combined with a "*" key:

```
        "components": ["./custom_app/beam/components/"],
	# (or alternatively)
	# "components": {"*": "./custom_app/beam/components/"},
	# "components": {"Repack": "./custom_app/beam/components/Repack.vue"},

```

agritheory:
@Alchez Got in some time on this this afternoon with the following WIP:
- hooks resolve in the apps resolution order as specified in apps.json
- Moved route and component config to hooks.py so the can be more easily overridden in config by client apps as needed
- Builds successfully but without importing the components or routes. That's based on [this work](https://uvr.esm.is/guide/extending-routes) that I didn't completely understand but seemed like a promising route
- Typing needs some attention, TS is not a fan of me saying "it's an object, what's the problem"
- We want the `/home` page to be configurable in the UI and role restricted, but that's a future step built on this work

.gitignore
```diff
@@ -197,7 +197,12 @@ cypress/videos
...
-
 beam/www/beam/index.js
 beam/www/beam/index.js.map
 beam/www/beam/index.css
+
+# unplugin-vue-components outputs
+beam/www/beam/components.d.ts
+
+# unplugin-vue-router outputs
+beam/www/beam/typed-router.d.ts
```
Alchez, line 208: Although the plugins recommend committing these files, I think we should ignore them since they will be statically built per bench folder and will be different for different clients.

beam/hooks.py
```diff
@@ -516,3 +516,88 @@
...
+		"Workstation": "./beam/beam/www/beam/pages/Workstation.vue",
+		"WorkOrder": "./beam/beam/www/beam/pages/WorkOrder.vue",
+		"Receive": "./beam/beam/www/beam/pages/Receive.vue",
+		"Ship": "./beam/beam/www/beam/pages/Ship.vue",
+		"Demand": "./beam/beam/www/beam/pages/Demand.vue",
+		"Manufacture": "./beam/beam/www/beam/pages/Manufacture.vue",
+		"Move": "./beam/beam/www/beam/pages/Move.vue",
+		"Repack": "./beam/beam/www/beam/pages/Repack.vue",
+		"JobCard": "./beam/beam/www/beam/pages/JobCard.vue",
+		"Operation": "./beam/beam/www/beam/pages/Operation.vue",
```
Alchez, line 533: @agritheory Since these paths will be analyzed statically during build-time and must still remain relative, I think we should document somewhere what is expected in these hooks.

The current rule is: path from an app's root level-folder. This allows a custom app to re-use components from _any_ app.

What do you think?

.gitignore
agritheory: Agreed

beam/hooks.py
agritheory: Yes, I think that makes sense, both from a DX and documentation perspective

beam/www/beam/plugins/component.ts
```diff
@@ -0,0 +1,48 @@
...
+import { HookConfig } from '@/types/config.js'
+import { getAppConfigs, mergeConfigs } from './hooks.js'
+
+export function getComponentPluginOptions(): Options {
+	// let globs: Options['globs']
+	// const components = getComponents()
+	// if ('*' in components) {
+	// 	globs = components['*']
+	// 	delete components['*']
+	// }
```
Alchez, line 12: @agritheory I was trying to allow component discovery using wildcards, but I'm not sure how the plugin expects me to pass the directory values. I'll do some more research later.

PR #172, opened 2024-09-13
Transfer work order

### LINKED ISSUE #163: Transfer Stock / Manufacture in Work Order
For both the "Transfer for Manufacture" and "Manufacture" steps we need a a `ListView` component that can show the following data:
- Item Code:Item Name
- Source and Target Warehouses
- Transfer Qty/Required Qty

image

This will need to incorporate some other changes to be considered complete:
 - Metadata component #161 
 - Save/Submit/Cancel Stock Entry workflow #164

lauty95:
I created Transfer component which show `requiredItems` from data fetched from a Work Order

Captura de pantalla 2024-09-13 a la(s) 1 28 28 p  m


beam/www/beam/index.ts
```diff
@@ -71,9 +71,9 @@ const routes: RouteRecordRaw[] = [
 		meta: { requiresAuth: true },
 	},
 	{
-		path: '/transfer',
-		name: 'transfer',
-		component: Transfer,
+		path: '/demand',
+		name: 'demand',
+		component: Demand,
```
Alchez: I think it may have shown up from a bad merge, but this version without the `meta` information is safe to remove

beam/www/beam/pages/WorkOrder.vue
```diff
@@ -1,33 +1,46 @@
...
 import { useFetch } from '../fetch'
+import Transfer from '../components/Transfer.vue'
 
 const route = useRoute()
 const workOrderId = route.params.orderId
 
 const workOrder = ref<Partial<WorkOrder>>({})
 const operations = ref<ListViewItem[]>([])
 const jobCards = ref<ListViewItem[]>([])
+const requiredItems = ref<ListTransferItem[]>([])
```
Alchez: I don't think this is being used anywhere, but ignore if this is WIP

yarn.lock
Alchez, line 1: Since Frappe apps work off of Yarn, I think you might want to regenerate this using `yarn install` instead

Alchez (changes requested):

beam/www/beam/components/Transfer.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+    <div>
+        <table class="item-table">
```
agritheory: @lauty95 Can you refactor this to use ListView and ListItem

agritheory:
@lauty95 merged into mobile_v15_work_order_refactor

PR #180, opened 2024-09-25
Mobile v15 control buttons

lauty95:
Created Control Buttons component and handled buttons status based on doctstatus

### DIFF
--- a/beam/www/beam/components/ControlButtons.vue
+++ b/beam/www/beam/components/ControlButtons.vue
@@ -1,21 +1,55 @@
-<!-- ControlButtons.vue -->
 <template>
 	<div class="control-buttons">
-		<button @click="onCreate" :disabled="!!docstatus">Save</button>
-		<button @click="onSubmit" :disabled="!docstatus">Submit</button>
-		<button @click="onCancel" :disabled="!docstatus">Cancel</button>
+		<button @click="onCreate" :disabled="docstatus !== -1">Save</button>
+		<button @click="onSubmit" :disabled="docstatus !== 0">Submit</button>
+		<button @click="onCancel" :disabled="docstatus !== 1">Cancel</button>
 	</div>
 </template>
 
 <script setup lang="ts">
-import { defineProps } from 'vue'
+import { defineProps, ref } from 'vue'
+const docstatus = ref(-1)
 
 const props = defineProps<{
-	docstatus: number
-	onCreate: () => Promise<void>
-	onSubmit: () => Promise<void>
-	onCancel: () => Promise<void>
+	onCreate: () => Promise<{ data: any }>
+	onSubmit: () => Promise<{ data: any; exception: string; response: Response }>
+	onCancel: () => Promise<{ data: any; exception: string; response: Response }>
 }>()
+
+const onCreate = async () => {
+	try {
+		const { data } = await props.onCreate()
+		if (data) {
+			docstatus.value = data.docstatus
+		}
+	} catch (err) {
+		console.error(err)
+	}
+}
+
+const onSubmit = async () => {
+	try {
+		const { data } = await props.onSubmit()
+		if (data) {
+			docstatus.value = data.docstatus
+		}
+	} catch (err) {
+		console.error(err)
+	}
+}
+
+const onCancel = async () => {
+	try {
+		const { data } = await props.onCancel()
+		if (data) {
+			docstatus.value = -1
+		}
+	} catch (err) {
+		console.error(err)
+	}
+}
+
+
 </script>
 
 <style scoped>
--- a/beam/www/beam/components/ControlButtons.vue
+++ b/beam/www/beam/components/ControlButtons.vue
@@ -1,19 +1,20 @@
 <template>
 	<div class="control-buttons">
-		<button @click="onCreate" :disabled="docstatus !== -1">Save</button>
-		<button @click="onSubmit" :disabled="docstatus !== 0">Submit</button>
-		<button @click="onCancel" :disabled="docstatus !== 1">Cancel</button>
+		<button class="btn" @click="onCreate" :disabled="docstatus !== -1">Save</button>
+		<button class="btn" @click="onSubmit" :disabled="docstatus !== 0">Submit</button>
+		<button class="btn" @click="onCancel" :disabled="docstatus !== 1">Cancel</button>
 	</div>
 </template>
 
-<script setup lang="ts">
+<script setup lang="ts" generic="T extends ParentDoctype">
+import { ParentDoctype } from '@/types';
 import { defineProps, ref } from 'vue'
 const docstatus = ref(-1)
 
 const props = defineProps<{
-	onCreate: () => Promise<{ data: any }>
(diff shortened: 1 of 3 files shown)


beam/www/beam/components/ControlButtons.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+	<div class="control-buttons">
+		<button @click="onCreate" :disabled="docstatus !== -1">Save</button>
```
agritheory: Can you match the styles from the navbar button?

beam/www/beam/types/index.ts
```diff
@@ -91,6 +91,7 @@ export type JobCard = ParentDoctype & {
 
 export type StockEntry = ParentDoctype & {
 	items: WorkOrderItem[]
+	docstatus: number
```
Alchez: This can be a part of the `ParentDoctype` type interface since it applies to all doctypes

beam/www/beam/components/Transfer.vue
```diff
@@ -24,7 +27,7 @@ const store = useDataStore()
 const listItems = ref<ListViewItem[]>([])
 const stockEntryId = ref('')
 
-const create = async () => {
+const create = async (): Promise<{ data: any }> => {
```
Alchez: This should already be typed to return a `StockEntry` object so I don't think you need an explicit response:

image

beam/www/beam/components/ControlButtons.vue
```diff
@@ -0,0 +1,61 @@
...
+		<button @click="onCancel" :disabled="docstatus !== 1">Cancel</button>
+	</div>
+</template>
+
+<script setup lang="ts">
+import { defineProps, ref } from 'vue'
+const docstatus = ref(-1)
+
+const props = defineProps<{
+	onCreate: () => Promise<{ data: any }>
```
Alchez: I think this button will involve both create and update actions. Maybe rename this to `onSave` instead?

beam/www/beam/components/ControlButtons.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+	<div class="control-buttons">
+		<button @click="onCreate" :disabled="docstatus !== -1">Save</button>
+		<button @click="onSubmit" :disabled="docstatus !== 0">Submit</button>
+		<button @click="onCancel" :disabled="docstatus !== 1">Cancel</button>
+	</div>
+</template>
+
+<script setup lang="ts">
```
Alchez: To avoid having to use `any` everywhere in the component and add additional typing in the IDE, you could add a [generic type](https://vuejs.org/api/sfc-script-setup#generics) here. Something like:

```suggestion
<script setup lang="ts" generic="T extends ParentDoctype">
```

That way, your `props` can change to:

```ts
const props = defineProps<{
	onCreate: () => Promise<{ data: T }>
	onSubmit: () => Promise<{ data: T; exception: string; response: Response }>
	onCancel: () => Promise<{ data: T; exception: string; response: Response }>
}>()
```

And you automatically get the following in the IDE:

image

beam/www/beam/components/ControlButtons.vue
Alchez: The `docstatus` checks will probably have to change to account for Frappe's enum values and conditions:

```mermaid
stateDiagram-v2
	state "Draft/Unsaved (0)" as draft
	state "Submitted (1)" as submitted
	state "Cancelled (2)" as cancelled

    [*] --> draft

    draft --> submitted: Submit
    submitted --> cancelled: Cancel
```

beam/www/beam/components/Transfer.vue
```diff
@@ -34,6 +37,7 @@ const create = async () => {
 	if (data.name) {
 		stockEntryId.value = data.name
 	}
+	return { data }
```
Alchez: I think this should return the other values from the `insert` call as well (`exception`, `response`).

Alchez (changes requested):

agritheory:
@lauty95 Can you mark this ready when you've addressed all the review comments? It looks like you're close

lauty95:
> @lauty95 Can you mark this ready when you've addressed all the review comments? It looks like you're close

yea, sure!

beam/www/beam/components/ControlButtons.vue
lauty95: Hey @Alchez! Thanks for you review.

I just updated these buttons. Could you tell me if it's ok that way?

lauty95:
@agritheory
I added amend button and change the disabled status on buttons based on docstatus and name of doctype.

I'll await for Rohan's review

beam/www/beam/components/Transfer.vue
```diff
@@ -13,6 +13,7 @@ import { ref } from 'vue'
 
 import { useDataStore } from '@/store'
 import type { ListViewItem, StockEntry } from '@/types'
+import ControlButtons from '../components/ControlButtons.vue'
```
Alchez: Minor nitpick, but a nicer way to shorthand this is with TS' path configs, so this can become:

```suggestion
import ControlButtons from '@/components/ControlButtons.vue'
```

beam/www/beam/components/ControlButtons.vue
```diff
@@ -0,0 +1,65 @@
...
+	onCreate: () => Promise<{ data: T }>
+	onSubmit: () => Promise<{ data: T; exception: string; response: Response }>
+	onCancel: () => Promise<{ data: T; exception: string; response: Response }>
+}>()
+
+const onSave = async () => {
+	try {
+		const { data } = await props.onCreate()
+		doctypeName.value = data.name
+		if (data) docstatus.value = data.docstatus
```
Alchez: This needs to be wrapped around the same if block:

```suggestion
		if (data) {
			doctypeName.value = data.name
			docstatus.value = data.docstatus
		}
```

beam/www/beam/components/ControlButtons.vue
```diff
@@ -0,0 +1,65 @@
...
+</template>
+
+<script setup lang="ts" generic="T extends ParentDoctype">
+import { ParentDoctype } from '@/types';
+import { defineProps, ref } from 'vue'
+const docstatus = ref(0)
+const doctypeName = ref("")
+
+const props = defineProps<{
+	onCreate: () => Promise<{ data: T }>
```
Alchez: Add exception and response to the types too.

Optionally, this could be simplified into a type as well. Something like:

```ts
export type DocActionResponse<T> = {
	data: T
	exception: string
	response: Response
}
```

Then the prop types can just be:

```ts
onCreate: () => Promise<DocActionResponse<T>>
```

Alchez (changes requested):
Hey @lauty95, sorry for the late reply. I think there's some minor changes that can happen here, but otherwise this is looking good.

lauty95:
> Hey @lauty95, sorry for the late reply. I think there's some minor changes that can happen here, but otherwise this is looking good.

Perfect!

Alchez (approved):

PR #184, opened 2024-09-27
feat: implement pagination for demand

### LINKED ISSUE #160: [demand] Add pagination to `get_demand` API

Alchez:
Closes #160 

---

**Changes:**
- Apply pagination to the Demand component using VueUse's [`useInfiniteScroll`](https://vueuse.org/core/useInfiniteScroll/#useinfinitescroll) composable
- Cleanup data initialization in some components via the Pinia data store
- Add arguments to existing `getAll` method for paginating Frappe's resource API responses
- Change Vite's build output format from ES to UMD to mitigate global namespace clash issues (also creates a smaller output but a larger sourcemap)

### DIFF
--- a/beam/www/beam/pages/Demand.vue
+++ b/beam/www/beam/pages/Demand.vue
@@ -7,12 +7,14 @@
 			<RouterLink :to="{ name: 'home' }">Home</RouterLink>
 		</template>
 	</Navbar>
+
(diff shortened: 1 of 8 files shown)


agritheory (approved):
Good find with the off-the-shelf composable

PR #185, opened 2024-09-27
[mobile / demand] Add accounting dimensions to Demand DB and allow filtering by those dimensions in get_demand API 

fproldan:
https://github.com/agritheory/beam/issues/168


beam/beam/demand/demand.py
```diff
@@ -633,33 +633,24 @@ def get_descendant_warehouses(company: str, warehouse: str) -> list[str]:
 
 
 @frappe.whitelist()
```
agritheory, line 635: Can we match the keywods/API of `frappe.get_all`?

agritheory (approved):

Issue #157, opened 2024-08-16
[demand] Calculate demand for Receiving and Shipping contexts


agritheory:
### Receiving 

Aggregating demand signals of Purchase Order, ~~Blanket Orders~~, Subcontracting Order and un-received Purchase Invoices without a Purchase Order, compute a new table to be added to the demand sqlite db. This should only account for inventoriable items. This should be available at `get_receiving_demand` endpoint that matches the `get_demand` function signature.

We want to be able to configure separately: 
Where a Purchase Order has been generated from a Sales Order or Work Order (via subcontracting workflow), create an allocation for it even if the inventory has not yet been received. 

### Shipping

We want to be able to signal when a Sales Order Item has been satisfied and move that demand from the Demand/Allocation tables into a separate one that mirrors the purchasing table above. This should be available at `get_shipping_demand` that matches the `get_demand` function signature.

PR #188, opened 2024-09-28
Receiving demand

agritheory:
https://github.com/agritheory/beam/issues/157


agritheory:
@fproldan I was talking through this feature with our customer and the interaction with drop ship POs came up. We should be ignoring those, but I don't think we need to include a test for that.

agritheory:
@fproldan I am going to merge this as WIP into `mobile_v15` so we can continue with front end work on it. Please add the additional tests planned/discussed in another PR

Issue #200, opened 2024-10-10
[demand] Escape special characters in item name


agritheory:
Item code `Repair Tape - 2" x 50'` throws an error. 

I think it probably makes the most sense to refactor to pypika for the queries and use sqlite for creating the schema.

PR #203, opened 2024-10-11
[demand] Escape special characters in item name

fproldan:
https://github.com/agritheory/beam/issues/200

### DIFF
--- a/beam/beam/demand/demand.py
+++ b/beam/beam/demand/demand.py
@@ -684,13 +684,16 @@ def get_demand(*args, **kwargs) -> list[Demand]:
 	demand = Table("demand")
 	allocation = Table("allocation")
 
-	d_filters = a_filters = []
+	d_filters = []
+	a_filters = []
+
 	if kwargs.get("filters"):
-		filters = kwargs.get("filters")
-		if filters:
-			for key, value in filters.items():
-				d_filters.append(demand[key].isin(value))
-				a_filters.append(allocation[key].isin(value))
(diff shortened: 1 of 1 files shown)


beam/beam/demand/demand.py
```diff
@@ -632,106 +676,137 @@ def get_descendant_warehouses(company: str, warehouse: str) -> list[str]:
...
-		ORDER BY delivery_date, idx, creation, parent ASC
-	"""
+	if a_filters:
+		allocation_query = allocation_query.where(*a_filters)
 
 	record_offset = records_per_page * (page - 1)
-	query = (
-		f"{demand_query} UNION ALL {allocation_query} LIMIT {records_per_page} OFFSET {record_offset}"
-	)
+	final_query = demand_query.union(allocation_query).limit(records_per_page).offset(record_offset)
```
fproldan: TO FIX: individual queries works, the union doesn't

agritheory: @fproldan Heather figured out a strategy for this in Check Run where we needed some queries to be configurable

https://github.com/agritheory/check_run/blob/version-14/check_run/check_run/doctype/check_run/check_run.py#L702-L719

fproldan:
The tests are passing locally

Captura de pantalla 2024-10-11 a la(s) 2 27 34 p  m

agritheory (approved):
@fproldan This looks great! Demand map builds with the customer data where we discovered the problem.

PR #208, opened 2024-10-15
feat: add config to ignore drop shipped items in demand

### LINKED ISSUE #183: [demand] Add configuration to ignore demand for items that are to be drop shipped

HKuz:
Closes #183 

Refactored the PR to branch off the `receiving_demand` branch to leverage that branch's queries.


agritheory:
@HKuz Try using a dataclass to define the types for Demand and Allocation tables to defeat this mypy error

agritheory (changes requested):
There has to be another way to do this, bracket syntax is terrible to read.
Try it this way: https://github.com/frappe/frappe/blob/develop/frappe/__init__.py#L56

HKuz:
@agritheory okay - I'll try maybe a mixin of TypedDict and _dict, but that was the core of most of the issues: the TypedDict object didn't support dot notation. (We may want to tighten some of the allowed types for the keys, too - the most recent set of errors were because a key could be a str or float, then we try to add/subtract/etc with it and mypy complained. I can get rid of the errors with casting, but that's also ugly.

agritheory (approved):

Issue #201, opened 2024-10-10
[mobile] Detect mobile user agents on login and redirect to `/beam`


agritheory:
Let's try this with a Role-style redirect , "BEAM User"

lauty95:
@agritheory Question..
If you have a `BEAM User` role, do you always have to be redirect to /beam? This means that a Beam user role will never have access to /app
What if you haven't `BEAM User` role? In this case if you are in mobile you should be redirect to /beam and if you are in desktop/notebook, etc you should be redirect to /app?

Am I right?

agritheory:
> If you have a BEAM User role, do you always have to be redirect to /beam? This means that a Beam user role will never have access to /app

This is desirable and reflects the mobile user experience we're looking for

> What if you haven't BEAM User role? In this case if you are in mobile you should be redirect to /beam and if you are in desktop/notebook, etc you should be redirect to /app?

I think it's fine for a desktop user to navigate to /beam intentionally and use it.

agritheory:
This currently redirects a user who does not have the BEAM User Role to `/beam` which is not desirable. We want the role but I think the user agent detection isn't going to help like I thought- all Firefox users are detected as Android mobile

agritheory:
Resolved inside of #248

PR #212, opened 2024-10-31
Detect mobile user agents on login and redirect to /beam

fproldan:
https://github.com/agritheory/beam/issues/201


fproldan:
@agritheory 

I figured out how to do the redirection directly using user agents as stated in the original issue (`auth_hooks` + overriding `home_page`). If you still want to choose to redirect based on a role I can make the change, only we need to change the condition of the redirection.

agritheory (changes requested):
This looks great but we want another logical check in both boot.py and the web bundle - that the user has the role "BEAM Mobile User". Similar to the "Supplier" and "Customer" roles, I think we want to create this without Desk Access but it probably won't matter because they're likely to have other (desk) roles too.  

image

PR #217, opened 2024-11-08
feat: track unsaved changes in data store

### LINKED ISSUE #216: [mobile] compute dirty / unsaved state on record when item is scanned or entered
TBD if changes should be cached or discarded when the user navigates away from the page

I propose that a hash of the values in object are stored (in the store) when it is fetched from the server and then a computed can reference that same value at any time later. I think this should get a larger design discussion when implemented in @stonecrop/stonecrop.

Alchez:
Work towards #216 

---

This implements the [`store.$subscribe`](https://pinia.vuejs.org/core-concepts/state.html#Subscribing-to-the-state) functionality to track unsaved changes. As soon as a document is created, saved, submitted, or cancelled, the tracker will toggle off until the next set of changes.

**Unhandled edge cases:**
- Currently the tracker only works in form views and so the state management is setup against the active form in the store. I'm not sure if this assumption is fully correct though, and if the tracker needs to work in other contexts as well.
- If a user makes a change and reverts it, the tracker will still count it as an unsaved change. For deeper state comparisons, I think we'll have to look at other alternatives as described in https://github.com/agritheory/beam/issues/216#issuecomment-2462018299.


beam/www/beam/pages/WorkOrder.vue
```diff
@@ -2,14 +2,15 @@
 	<Navbar>
 		<template #title>
 			<h1 class="nav-title">Manufacture</h1>
+			<span v-if="store.form.dirty" class="nav-subtitle">Unsaved changes</span>
```
Alchez, line 5: @agritheory I've added an empty `nav-subtitle` class reference in case this needs to be styled. Currently, if any changes are made to the form, the page would look something like this:

image

agritheory (approved):

PR #219, opened 2024-11-12
Mobile v15 work order

### LINKED ISSUE #224: [mobile] Add hooks for scanning Handling Units in Work Order

agritheory:
To do:
- [x] Work Order needs to hold data for two or more other doctypes: the work order, mapped stock entries and connected operations
- [x] The Items ListView should reflect both the required qty and the unsaved "to transfer/manufacture" qty that will be mapped to the stock entry.
- [x] Scanning an item should take the appropriate action on the ListView (mapped to `items` of the mapped Stock Entry).
- [x] Other work started, the Purchase Receipt and Delivery Note documents are built and can be iterated to completion


beam/hooks.py
```diff
@@ -544,19 +546,19 @@
...
 		{
-			"path": "/job_card/:orderId/",
+			"path": "/job_card/:id/",
 			"name": "job_card",
 			"component": "JobCard",
 			"meta": {"requiresAuth": True, "doctype": "Work Order", "view": "form"},
 		},
 		{
-			"path": "/work_order/:orderId/operation/:id",
+			"path": "/work_order/:id/operation/:id",
```
Alchez: @agritheory The route params will be resolved in an object using the names you provide in the path, so I think this will cause a regression since you can't have two identical keys in an object: `{ id: string, id: string }`

beam/www/beam/scan.ts
```diff
@@ -47,8 +47,9 @@ export class ScanHandler {
 
 	add_or_associate(barcode_context: FormContext[]) {
 		barcode_context.forEach(async action => {
+			const parentfield = action.parentfield || 'items'
 			const is_stock_entry =
-				this.store.form.doctype === 'Stock Entry' &&
+				(this.store.form.doctype === 'Stock Entry' || 'Work Order') &&
```
Alchez: @agritheory I don't think the string change (+ 'Work Order') will do anything since both conditions in this check need to be satisfied for the variable to be truthy, and the second check will never be true for Work Order since the `stock_entry_type` field doesn't exist on WO.

Alchez (changes requested):

beam/hooks.py
agritheory: Got it, order is ugly, but what I guess we'll have to use.

beam/www/beam/scan.ts
agritheory: I was trying to unblock the reactivity of the items table via scan, but I think you've gotten that figured out elsewhere.

agritheory:
@Alchez I've tested the WIP here and I think we need a slight adjustment in the workflow. After the user has saved the stock entry, only those items appear in the ListView and none of the other ones do. I think this makes even a brief pause-and-resume pattern frustrating. The functionality we're looking for is that the 'transfer_qty' field belongs to the SE, the 'required_qty' field belongs to the WO and the rest of the data - which is not editable - can be common and copied over to the server.

Alchez:
@agritheory Okay, this should now create and update draft Stock Entries in the Manufacture tab correctly. Can you give it a test?

Issue #215, opened 2024-11-05
[mobile] Refactor control buttons


agritheory:
These workflows should exist in the appropriate components and they should pass both a label and callback to the control buttons component. 

## Manufacture | Work Order

```mermaid
flowchart TD

A[Work Order] --[Skip Transfer is True] --> B["'Save', 'Manufacture'"]
A --[Skip Transfer is False] --> C["'Save', 'Transfer'"]
C -- [Transfer for Manufacture SE exists] -->  D["'Save', 'Manufacture', 'Cancel Transfer'"]
B -- [Manufacture SE exists for != 100% of required items] --> E["'Complete', 'Cancel Manufacture'"]
B -- [Manufacture SE exists for == 100% of required items] --> F["'Cancel Manufacture'"]
D -- [Manufacture SE exists for != 100% of required items] --> E["'Complete', 'Cancel Manufacture'"]
D -- [Manufacture SE exists for == 100% of required items] --> F["'Cancel Manufacture'"]
```

When Work Order is marked as complete, Work Order is closed, allowing for less than 100% manufacture.

## Receive | Purchase Receipt
```mermaid
flowchart TD

L[Purchasing Document] -- [Purchase Receipt(s) exist for < 100% of items] --> M["'Save', 'Receive'"]
```

## Ship | Delivery Note 
```mermaid
flowchart TD

R[Sales Document] -- [Delivery Note(s) exist for < 100% of items] --> S["'Save', 'Ship'"]
```

agritheory:
```
[{
"label": "Save",
"action":  store.save()
},
{
"label": "Ship",
"action":  store.submit()
}]
```

PR #222, opened 2024-11-13
refactor: control buttons

lauty95:
WIP
An idea of use the ControlButtons component based on issue #215 

```vue
<template>
    <ControlButtons :buttons="controlButtons" />
</template>
<script>
const controlButtons = computed(() => {
	return [
		{
			label: 'Save',
			action: create,
			disabled: listItems.value.length === 0,
		},
		{
			label: 'Submit',
			action: () => store.submit<StockEntry>('Stock Entry', stockEntryId.value),
			disabled: stockEntryId.value === '',
		},
		{
			label: 'Cancel',
			action: () => store.cancel<StockEntry>('Stock Entry', stockEntryId.value),
			disabled: stockEntryId.value === '',
		},
	]
})
</script>
```

i have to work on styles, but based on the length of array the buttons should fit in its place

Captura de pantalla 2024-11-13 a la(s) 11 41 29 a  m
Captura de pantalla 2024-11-13 a la(s) 11 41 57 a  m
Captura de pantalla 2024-11-13 a la(s) 11 44 07 a  m


beam/www/beam/components/ControlButtons.vue
```diff
@@ -1,78 +1,40 @@
...
+	justify-content: flex-end;
+}
+
+.button-2 {
+	justify-content: space-between;
+}
+
+.button-3 {
+	justify-content: space-between;
+}
```
Alchez, line 39: How would this work for more than 3 buttons? Should these be made dynamic using parent flex justify values?

lauty95: Ignore my custom classes haha, just did that way to make some photos.

But I'll think about it for any number of buttons

agritheory: @lauty95 I think we don't need to support more than three. I'd be OK with stacking if we have three or four.

agritheory:
@lauty95 Can you add a "hidden" API in addition to "disabled"? I can think of a use case for each in the workflow we've put together already. Hidden should evaluate first and override disabled.

agritheory:
This was super easy: 
```
		<BeamBtn
			v-for="(button, index) in buttons"
			:key="index"
			@click="button.action"
			:disabled="button.disabled"
			v-show="!button.hidden"
		>
		```

agritheory (approved):
Great work. Merging even though it's in WIP

PR #226, opened 2024-11-20
wip: better import for third party apps

agritheory:
This PR provides two aliases - one to the beam/www/beam directory where components and store are located, and one for BEAM's node modules so that things like the Vue dependencies are directly available. If you want to import another dependency included in the customer app's package JSON, that's easy too.

```
// third party app .../thing.js
import { computed, ref } from "@beamNode/vue";
import { useRoute } from "@beamNode/vue-router";

import ControlButtons from '@beam/components/ControlButtons.vue'
import { useBeamStore } from '@beam/stores/beam'

const route = useRoute()
const store = useBeamStore()
```


Alchez:
@agritheory I'm not sure about this change. Is the intention to use the Beam store (pinia) in third-party apps without it being initialized in the Vue instance?

PS: externalizing the dependencies seems to stop the base Beam instance from rendering at all for me

PR #237, opened 2024-11-27
test: add playwright tests

### LINKED ISSUE #193: [mobile] Test Suite
Use [`pytest-playwright`](https://github.com/agritheory/approvals/blob/customer-credit-limit/approvals/tests/test_non_workflow_approval.py) to test the basic features (MVP)

## Receive | Purchase Receipt
- [x] scan item barcode to receive item (this is a proxy for a UPC scan) 
- [x] complete partial receipt
- [ ]  cancel PR
- [ ] complete a receipt with all items 

## Manufacture | Work Order
- [x] scan an item to increment
- [ ] scan a handling unit to add
- [ ] complete transfer for mfg
- [ ] complete job cards
- [ ] complete mfg SE from work order
- [ ] "short" a mfg SE
- [ ] "complete" a mfg SE

## Ship | Delivery Note 
- [x] scan item barcode to increment
- [ ] scan a handling unit to add
- [ ] complete partial shipment against SO
- [ ]  cancel DN
- [ ] complete DN with all items

---

Out of scope:
- Extension APIs used by other client applications
- Login / out / redirect (will be tested passively as part of other tests)

Alchez:
Closes #193 

---

Test locally inside a virtualenv using `pytest ./beam/tests/mobile/test_mobile.py --browser firefox --headed --disable-warnings`


beam/tests/test_mobile.py
```diff
@@ -0,0 +1,65 @@
...
+
+@pytest.mark.parametrize("route", ["Receive", "Ship", "Manufacture"])
+def test_scan_item_barcode(page, route):
+	# navigate in the following order: Home -> List -> Form
+	page.get_by_text(route).click()
+	page.locator("css=.beam_list-item").first.click()
+
+	# find the first item in the list
+	item = page.locator("css=.box .beam_list-item").first
+	item_name, *others = item.inner_text().split("\n")
```
Alchez: @agritheory this only works cause the first thing in each list is the item name, but we could maybe tag the item name with a class to identify it more reliably in tests?

agritheory: I'd prefer a data attribute, but I agree.

Alchez:
@agritheory I'm seeing some weird behaviour that I'm not sure if you've faced before. While running the tests locally, I'm scanning and saving the record and it creates an entry in the database. However, the pytest runner can't seem to fetch that record for assertions.

Have you seen something like this?

fproldan:
@agritheory 


Some things I tried:

- Commit after SAVE
```python
page.get_by_text("SAVE", exact=True).click() 
frappe.db.commit()
```

- Time after SAVE
```python
page.get_by_text("SAVE", exact=True).click()
time.sleep(2)
```

- Scope function in db_instance()
```python
@pytest.fixture(scope="function", autouse=True) def db_instance():
```

- Re-connect after SAVE
```python
page.get_by_text("SAVE", exact=True).click() 
currentsite = "test_site"
sites = Path(get_bench_path()) / "sites"
if (sites / "common_site_config.json").is_file():
    currentsite = json.loads((sites / "common_site_config.json").read_text()).get("default_site")

frappe.init(site=currentsite, sites_path=sites, force=True)
frappe.connect()
```

- Disabled `delete_draft_records` temporarily to see if it is interfering.
- Check that the URL from `setup` is the same that the URL used by Playwright.
- Combinations of the above.

lauty95:
I found that when the browser's transaction ends, the data appears when I use `bench console`.

After researching and debugging, I noticed that the test code and the browser work as two different users consulting the same database but operating in two different transaction "snapshots". When the browser makes API calls and saves data via HTTP requests, the test code needs to refresh its transaction to see the recently committed data.

```py
frappe.db.rollback()  # End current transaction (from browser)
frappe.db.begin()     # Start new transaction with updated snapshot
```

## Demo

https://github.com/user-attachments/assets/7d616f30-df8f-4c3f-86ae-e0a273cb1e5e

beam/tests/mobile/test_receive.py
```diff
@@ -0,0 +1,88 @@
...
+	):
+		page.evaluate("barcode => scanner.simulate(window, barcode)", barcodes[0])
+		expect(item_count).to_have_text(re.compile("1/"))
+
+	# ensure there are no existing Purchase Receipts against this Purchase Order
+	receipt = frappe.db.exists(
+		"Purchase Receipt Item",
+		{"docstatus": 0, "purchase_order": order_id, "item_code": item_code},
+	)
+	#assert not receipt
```
Alchez: How come we're removing this check?

beam/tests/mobile/test_receive.py
```diff
@@ -0,0 +1,88 @@
...
+	page.wait_for_timeout(1000)
+	frappe.db.rollback()
+	frappe.db.begin()
+	receipts = frappe.get_all(
+		"Purchase Receipt Item",
+		filters={"purchase_order": order_id, "item_code": item_code},
+		fields=["docstatus", "received_qty", "creation"],
+		order_by="creation desc",
+	)
+	assert len(receipts) >= 1
```
Alchez, line 196: I don't think we should be expecting more than 1 purchase receipt here. This should be an exact match to one receipt.

beam/tests/mobile/test_receive.py
```diff
@@ -0,0 +1,88 @@
...
+	page.get_by_text("RECEIVE", exact=True).click()
+	page.wait_for_timeout(1000)
+	frappe.db.rollback()
+	frappe.db.begin()
+	receipts = frappe.get_all(
+		"Purchase Receipt Item",
+		filters={"purchase_order": order_id, "item_code": item_code},
+		fields=["docstatus", "received_qty", "creation"],
+		order_by="creation desc",
+		limit=1
```
Alchez: Limiting the response to just one entry feels like hacking the test to fit the response. We should check that there's still just one receipt and that it's submitted.

Alchez (commented):
@lauty95 That's a great find, but also so weird. How is it even spawning multiple user connections in the test environment?

lauty95:
@Alchez Sorry if I'm not clear - this isn't my strongest area and my responses are based on some research + using print statements for debugging 😅

It's not actually "multiple user connections" - it's the same database connection but different transaction scopes.

**Flow breakdown:**

1. **Pytest starts** > Transaction A begins here:
   https://github.com/agritheory/beam/blob/7ed1c82f201fa7782795cba2e63943a01cb0faa2/beam/tests/mobile/test_receive.py#L18
   `test_complete_partial_receipt > conftest.py > db_instance > frappe.connect()`
   https://github.com/agritheory/beam/blob/7ed1c82f201fa7782795cba2e63943a01cb0faa2/beam/tests/conftest.py#L45

2. **Browser clicks** > HTTP POST to Frappe server:
   https://github.com/agritheory/beam/blob/7ed1c82f201fa7782795cba2e63943a01cb0faa2/beam/tests/mobile/test_receive.py#L60

3. **Frappe server** > Transaction B: BEGIN > INSERT > COMMIT > END
   (Server creates new transaction for the HTTP request, saves data, commits immediately)

4. **Pytest queries** > Still in Transaction A - can't see Transaction B's commits:
   https://github.com/agritheory/beam/blob/7ed1c82f201fa7782795cba2e63943a01cb0faa2/beam/tests/mobile/test_receive.py#L64
   (Returns empty because Transaction A sees old snapshot)

**Solution:**
```python
frappe.db.rollback()  # End Transaction A
frappe.db.begin()     # Start a new transaction with fresh snapshot
receipts = frappe.get_all(...)  # Now sees data from Transaction B
```

agritheory:
I think we need a helper function here to explicitly acquire the current transaction scope. I'd personally want to use an API that is one of a context manager or a decorator to be clear I was doing this. This would require no changes in other tests.

```python
def test_ui_thing(example_pytest_fixture):
    with use_current_db_transaction():
      value = frappe.get_value('Doctype', ...)
      assert value == True
```
It may have to be like this:

```python
def test_ui_thing(example_pytest_fixture):
    with use_current_db_transaction() as _db:
      value = _db.get_value('Doctype', ...)
      assert value == True
```

```python
@use_current_db_transaction()
def test_ui_thing(example_pytest_fixture):
   value = frappe.get_value('Doctype', ...)
   assert value == True
````

Or just a function call, which is to me, least clear.

```python

def test_ui_thing(example_pytest_fixture):
    use_current_transaction()
    value = frappe.get_value('Doctype', ...)
   assert value == True
```

The concern with using a decorator is that ordering matters, it can have unintended side affects and it's the least explicit/obvious/clear when reading and reviewing code. I'm slightly in favor of the context manager API.

I'd like to prototype this code here and eventually move it to test_utils.

beam/tests/mobile/test_receive.py
lauty95: When we run bench execute 'beam.tests.setup.before_test' we're creating **Purchase Receipt** that contains Cloudberry. That code tests 'not receipt', but it's created by setup.py

beam/tests/mobile/test_receive.py
lauty95: If I'm not wrong, here the tests create a Draft Purchase Receipt https://github.com/agritheory/beam/blob/dfc532bfb3973563b6e968022c7453eeb38488cc/beam/tests/setup.py#L671

So we have 2 receipts instead of 1
Screenshot 2025-07-02 at 10 34 13 AM

beam/tests/mobile/test_receive.py
lauty95: That's true! As I answered before, the `before_test` function in setup.py creates draft Purchase Receipt. So by limiting to 1 and ordering the list descending, I'm trying to get the most recently created Purchase Receipt

lauty95:
@agritheory @Alchez I guess this is ready

PR #241, opened 2024-12-06
Mobile v15 move page

### LINKED ISSUE #220: [mobile] Move page functionality
This page is intended to work as an ad-hoc solution to moving inventory, outside of normal putaway and transfer/manufacturing operations.
The page has three primary interfaces:
- Source Warehouse
- Target Warehouse 
- Items list

For the Items list, it can start empty but be populated by scanning 
For the Source Warehouse and Target Warehouse, these should normally be populated by scanning but need to fallback to a link component. When a warehouse is scanned it should populate whichever of the fields is not already populated, with the source first if they are both empty. In the normal Frappe/ERPNext context a link has a "follow" icon on the right but in the mobile context this isn't useful and should be replaced with a "clear" button, denoted by the multiplication character.

lauty95:
### Move Page - WIP

I implemented **ADropdown** as a local component because I encountered some issues with the `@stonecrop/aform` component:

### Issues Found:
- **handleClickOutside**: This function clears my input whenever I click anywhere on the page
- **setResult**: This function calls `closeResults` without the `result` parameter. As a result, in the `closeResults` function, the `search` ref is empty, which causes the function to fail

```javascript
// ADropdown.vue in @stonecrop/aform
const closeResults = () => {
  isOpen.value = false;

  if (!items.includes(search.value)) { // Here search.value is still empty
    search.value = '';
  }
};
```

---

- I added some styles to the **ADropdown** component based on `@stonecrop/aform`
- I added a custom event in `beam-web.bundle.js`. I tried creating a new file for this, but I had some issues. For testing purposes, I worked directly in this file.

---

- I implemented **ADropdown** in the **Move** page.

Screenshot 2024-12-06 at 2 29 49 PM

### Notes:
1. I left the `@stonecrop/aform` package installed to show you how I initially tried to implement it in this project. If you spot an error in my approach, I can try to fix it.
2. If the changes I made to the **ADropdown.vue** component make sense, I could open a PR in `@stonecrop/aform` repository.

### TODO:

```javascript
// Move.vue
const handleScanned = (event: CustomEvent) => {
	const scannedData = event.detail[0].context.doc.name
	if (!sourceWarehouse.value) {
		sourceWarehouse.value = scannedData
	} else if (!targetWarehouse.value) {
		targetWarehouse.value = scannedData
	}
	/* 
	TODO: else > fill listItems
	*/
	console.log('Warehouse scanned:', scannedData)
}
```


Alchez:
@lauty95 The changes here in ADropdown look good, you can copy them into `@stonecrop/aform` too.

lauty95:
### Move Page - Ready for review

**Custom Dropdown Component:**  
Replaced the dropdown from `@stonecrop/aform` with a locally defined `ADropdown` component due to issues with `handleClickOutside` and `closeResults` functions in the original package.

**Warehouse Selection:**  
- Implemented `Source Warehouse` and `Target Warehouse` fields using `ADropdown`.  
- Added a "clear" button (×) to reset each field individually.

**Items List:**  
- Integrated the `ListView` component to display scanned items.  
- Added a placeholder text ("Scan to Begin") when the list is empty.

**Save and Move Actions:**  
- Added `SAVE` and `MOVE` buttons:  
  - **SAVE** creates or updates a `Stock Entry` document.  
  - **MOVE** submits the document and clears the form for a new entry.  
- Buttons are disabled until all required fields (`source warehouse`, `target warehouse`, and `items`) are filled.

**Scanner Integration:**  
The scanner function directly updates the store.

**Demo:**  

https://github.com/user-attachments/assets/48f136fa-3515-477c-b22f-a251563177bd

**Notes:**  
The `@stonecrop/aform` package was removed as the custom dropdown resolves the encountered issues.

agritheory:
@lauty95 This looks great and the drop down component feels really good. Let's increase the font size to 150% to help the mobile user / small screen issue and we want to filter out group warehouses in the queries.

lauty95:
@agritheory Ready!

Added some styles and filtered group warehouses:
```javascript
const loadWarehouses = async () => {
  const warehouses = await store.getAll<Warehouse[]>('Warehouse', {
    filters: JSON.stringify([['is_group', '!=', '1']]),
  })
  warehouseList.value = warehouses.map(warehouse => warehouse.name)
}
```
Screenshot 2024-12-19 at 11 11 25 AM
Screenshot 2024-12-19 at 11 12 07 AM

beam/www/beam/pages/Move.vue
```diff
@@ -8,36 +7,81 @@
...
-const stockEntry = ref(store.cache.mappers[stockEntryId.value] as StockEntry)
+const stockEntry = computed(
+	(): StockEntry =>
+		(store.cache.mappers[''] as StockEntry) || {
+			name: '',
+			stock_entry_type: 'Material Transfer',
+			items: [],
+			from_warehouse: '',
+			to_warehouse: '',
+		}
```
Alchez, line 53: @lauty95 I don't think we need a mapper here, since there won't be a source document. Maybe you want to use the `store.makeNewDoc` function to create a Stock Entry from scratcg?

beam/www/beam/pages/Move.vue
```diff
@@ -8,36 +7,81 @@
...
+	})
+	warehouseList.value = warehouses.map(warehouse => warehouse.name)
+}
+
+const clearField = (field: 'from_warehouse' | 'to_warehouse') => {
+	store.$patch(state => {
+		const mapper = state.cache.mappers['']
+		if (mapper) mapper[field] = ''
+	})
+}
```
Alchez, line 79: @lauty95 This should probably just update the local reference to the Stock Entry object, and skip the mapper

beam/www/beam/components/ADropdown.vue
Alchez, line 1: @lauty95 @agritheory Should we move the delta into `@stonecrop/aform` directly?

Alchez (changes requested):

beam/www/beam/pages/Move.vue
lauty95: @Alchez Sorry for the delay.

The reason I use the mapper is that when I scan a warehouse, the scanner updates the store directly. 

This behavior is based on this line: https://github.com/agritheory/beam/blob/9dc460e82f3fa661c7d2ff9e49e700673aa01a61/beam/www/beam/stores/scan.ts#L219

In that code, the scanner updates the store.cache.mappers[documentId.value] with the scanned data. By using a computed property, I ensure that stockEntry stays reactive and automatically reflects any updates from the store.

Do you think it is ok that way?

beam/www/beam/components/ADropdown.vue
lauty95: I create this PR [#236](https://github.com/agritheory/stonecrop/pull/236)

beam/www/beam/pages/Move.vue
Alchez: Hmm, so there's maybe another approach to try here:
- Whenever a page loads, the app tries to set the form context in the `store.form` variable using an ID in the query.
- For the move page, this variable would be an empty object since it's not fetching an existing form.
- Instead of a mapper, we could manually create a new Stock Entry doc and set it back in the `store.form` variable.
- That would let the store functions to work correctly out of the box.

What do you think?

beam/www/beam/components/ADropdown.vue
Alchez: I added some comments on that PR, let me know

beam/www/beam/pages/Move.vue
lauty95: If I understand correctly, I think this is a good approach. I'll run some tests and tag you once I upload the changes.

beam/www/beam/components/ADropdown.vue
Alchez: @lauty95 that PR is now merged

lauty95: I delete ADropdown custom component and start using ADropdown from stonecrop/aform

beam/www/beam/pages/Move.vue
lauty95: @Alchez I really stunk with this topic haha!

I'll try to explain myself..

When you scan  a warehouse is the case that I explain in the top comment, the scanner updates store.cache.mapper..

And when you scan an item anywhere in the app that action runs "add_or_increment" function https://github.com/agritheory/beam/blob/3e0f4f9d0d5870b50b2d8a38cab1bdcc94df7712/beam/www/beam/stores/scan.ts#L98
wich also works with mappers because mappedDoc is defined that way https://github.com/agritheory/beam/blob/3e0f4f9d0d5870b50b2d8a38cab1bdcc94df7712/beam/www/beam/stores/scan.ts#L25

The issue I'm facing is that I don't know how to determine where the scanner was executed or how to populate store.form instead of store.cache.mappers

If you see a way to approach this, could you please guide me?

lauty95:
@agritheory I updated Beam.vue to apply font-family throughout the app.

I also added @stonecrop/aform to use the ADropdown component instead of the local one.

I'm currently waiting for Rohan's response regarding a request he made

agritheory (approved):
Excellent! I will add a ticket for tests to be completed after the repack page since they'll have a lot of overlap

beam/www/beam/pages/Move.vue
Alchez: @lauty95 Oh you're right, I think I'll need to take this design and tweak it to avoid needing a mapper for a scan. I'll track this for myself, but thanks!

PR #243, opened 2025-01-10
fix: added remaining_time_in_mins in job card

lauty95:
I added `remaining_time_in_mins` to enable setup execution before the test.

This change is based on the [merge](https://github.com/frappe/erpnext/compare/v15.48.1...v15.48.2)


agritheory:
@lauty95 Can you cherry-pick this for a verison-15 PR too? Thanks for the fix!

PR #244, opened 2025-01-13
Update scan.js

diniemasjuki:
fix error in erpnext 15 pos section

### DIFF
--- a/beam/public/js/scan/scan.js
+++ b/beam/public/js/scan/scan.js
@@ -1,4 +1,14 @@
-import onScan from 'onscan.js'
+if (!window.OnScan) {
+    import('onscan.js')
+        .then((OnScan) => {
+            OnScan.attachTo(document, {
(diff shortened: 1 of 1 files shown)


beam/public/js/scan/scan.js
```diff
@@ -1,4 +1,14 @@
-import onScan from 'onscan.js'
```
Alchez, line 1: @diniemasjuki Losing this import reference means that the handler below may not work anymore.

https://github.com/agritheory/beam/blob/6d162576438ee91b23781b96ce0d01d2653d6846/beam/public/js/scan/scan.js#L37

Alchez (changes requested):

beam/public/js/scan/scan.js
diniemasjuki: i see.. will look further into it

PR #245, opened 2025-01-21
Update scan.js

diniemasjuki:
fix error onScan already declared in version 15

### DIFF
--- a/beam/public/js/scan/scan.js
+++ b/beam/public/js/scan/scan.js
@@ -1,4 +1,12 @@
-import onScan from 'onscan.js'
+if (typeof onScan === 'undefined' || typeof onScan !== 'function') {
+	import('onscan.js')
+    .then(module => {
+      const onScan = module.default || module;
+      window.onScan = onScan; 
+    })
+    .catch(error => {      
+    });	
+}
 
 function waitForElement(selector) {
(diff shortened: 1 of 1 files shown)


Alchez:
@diniemasjuki can you share the issue that you're trying to solve and/or steps to replicate it? That'll help me test it out locally as well.

diniemasjuki:
> @diniemasjuki can you share the issue that you're trying to solve and/or steps to replicate it? That'll help me test it out locally as well. 

Thanks sir for reaching me out.
To simplify,  i try to fix this issue in our erpnext version 15-16.

Main issue,
OnScan has already been declared / missing.

Symptoms:
1. POS system becomes an error. Only load half of it.
2. Barcodes not automatically generated of the item group name are not products.
3. Some other error that i cant remember which is caused of the existing of the onScan module. Sometimes it is missing, sometimes its already been declared

agritheory:
@diniemasjuki I would prefer that we make the POS page compatible with the rest of BEAM - Handling Units and custom scannable doctypes. What would meet your needs beyond removing this error?

agritheory:
This PR seems to use version-15 as the base branch but it being made to the `version-14` branch. Please reopen with correct alignment.

PR #246, opened 2025-01-24
feat: allow re-routing after scan input

Alchez:
**Feature changes:**
- Add API for routing after scan input in existing Beam hook
  - I've added an additional `route` key in the hook definition that will leave the responsibility of defining the target route to the app itself, considering they have the context on the built routes.
  - @agritheory Currently the hook is setup to combine route actions across multiple apps, but ultimately they need to resolve to a single target route. Currently the client receives a list of route actions back and I'm picking the first hook in that list, but I'm not sure if it should instead be a single object that's updated by every app?

**Additional changes:**
- Update Stonecrop dependencies to latest
- <strike>Update test setup to exclude account numbers from account name</strike>
- <strike>@agritheory it seems like creating production plans is failing during test setup (on frappe 15.53.0 and erpnext 15.49.3). I'm not sure how to fix it, but here's the traceback:</strike>

<details>
<summary>Traceback</summary>

```python
  File "/frappe/frappe/commands/utils.py", line 269, in execute
    ret = frappe.get_attr(method)(*args, **kwargs)
  File "/beam/beam/tests/setup.py", line 59, in before_test
    create_test_data()
  File "/beam/beam/tests/setup.py", line 106, in create_test_data
    create_production_plan(settings, prod_plan_from_doc)
  File "/beam/beam/tests/setup.py", line 613, in create_production_plan
    raw_materials = get_items_for_material_requests(
  File "/frappe/frappe/utils/typing_validations.py", line 31, in wrapper
    return func(*args, **kwargs)
  File "/erpnext/erpnext/manufacturing/doctype/production_plan/production_plan.py", line 1513, in get_items_for_material_requests
    frappe.throw(
  File "/frappe/frappe/__init__.py", line 603, in throw
    msgprint(
  File "/frappe/frappe/__init__.py", line 568, in msgprint
    _raise_exception()
  File "/frappe/frappe/__init__.py", line 519, in _raise_exception
    raise exc
frappe.exceptions.ValidationError: Items to Manufacture are required to pull the Raw Materials associated with it.
```

</details>

### DIFF
--- a/beam/www/beam/stores/scan.ts
+++ b/beam/www/beam/stores/scan.ts
@@ -159,10 +159,9 @@ export const useScanStore = defineStore('scan', () => {
 	}
 
 	const route = (barcode_context: ListContext[]) => {
-		for (const action of barcode_context) {
-			store.router.push(action.route)
-			return // only route based on the first action
-		}
+		// only route based on the last action in hooks
+		const action = barcode_context && barcode_context.at(-1)
+		store.router.push(action.route)
 	}
 
 	const set_item_code_and_handling_unit = (barcode_context: FormContext[]) => {
--- a/beam/www/beam/stores/scan.ts
+++ b/beam/www/beam/stores/scan.ts
@@ -53,7 +53,7 @@ export const useScanStore = defineStore('scan', () => {
 				'Manufacture',
 			].includes((mappedDoc.value as StockEntry).stock_entry_type)
 
-		barcode_context.forEach(async action => {
+		for (const action of barcode_context) {
 			const existing_rows = mappedDoc.value.items.filter(row => {
 				if (is_stock_entry) {
 					return row.item_code === action.context.item_code || row.handling_unit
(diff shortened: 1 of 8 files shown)


agritheory:
> Currently the hook is setup to combine route actions across multiple apps, but ultimately they need to resolve to a single target route. Currently the client receives a list of route actions back and I'm picking the first hook in that list, but I'm not sure if it should instead be a single object that's updated by every app?

Yes, this needs to be refactored (in both the app context and the mobile context) - I've had that comment in there for more than a year now, let's fix it.

agritheory:
No issues with multiple reinstalls on this branch 
```beam 15.1.3
erpnext 15.49.3
frappe 15.53.0
hrms 15.38.2
```

beam/www/beam/stores/scan.ts
```diff
@@ -151,27 +151,30 @@ export const useScanStore = defineStore('scan', () => {
...
 	const filter = (barcode_context: ListContext[]) => {
 		// TODO: apply filters to listview; use store router
 	}
 
 	const route = (barcode_context: ListContext[]) => {
-		// TODO: re-route to formview; use store router
+		for (const action of barcode_context) {
+			store.router.push(action.route)
+			return // only route based on the first action
+		}
```
Alchez: Only execute the last route instead of the first

PR #248, opened 2025-01-27
Mobile v15 scan to login

### LINKED ISSUE #218: [mobile] Scan to log in
In this context three things have to be set:
- The User should have a configured Item Barcode
- The User should have IP address restriction(s)
- The User Should have the Role "BEAM Mobile User"

lauty95:
### Login System on Scan
I added a hook to handle login via scanning on the `/login` path.

I included *scan.js* in the *beam-web.bundle.js* to enable the scanner functionality on the `/login` path.

Here’s a preview of the changes:

https://github.com/user-attachments/assets/0b199a04-a758-4331-acf8-7806c85ec431

The next step is to work on implementing *IP address restrictions* for users, as specified in the requirements on #218 issue.


beam/beam/overrides/user.py
```diff
@@ -0,0 +1,11 @@
...
+# For license information, please see license.txt
+
+from frappe.core.doctype.user.user import User
+from beam.beam.barcodes import create_beam_barcode
+
+
+class BEAMUser(User):
+	def validate(self):
+		super().validate()
+		create_beam_barcode(self)
```
Alchez: Generally class overrides are meant for changing existing functions inside a controller class. If you want to extend the validation logic, you can just use the [`validate`](https://docs.frappe.io/framework/v14/user/en/basics/doctypes/controllers#controller-hooks) hook in [`doc_events`](https://docs.frappe.io/framework/user/en/python-api/hooks#crud-events).

beam/hooks.py
```diff
@@ -199,7 +200,7 @@
 # Overriding Methods
 # ------------------------------
 
-# override_whitelisted_methods = {"demand": "beam.beam..graphql_server"}
+override_whitelisted_methods = {"scan_login": "beam.beam.scan.user_login.scan_login"}
```
Alchez: This hook is for overriding methods in other apps, but if it's the same app, you can just have the `frappe.xcall` caller use the dotted path to that function (I'm adding another comment to the relevant place).

beam/public/js/scan/scan.js
```diff
@@ -66,18 +83,24 @@ class ScanHandler {
...
-			const context = this.reduceContext()
-			frappe.xcall('beam.beam.scan.scan', { barcode: sCode, context: context, current_qty: iQty }).then(r => {
-				if (r && r.length) {
-					if (Object.keys(frappe.boot.beam.client).includes(r[0].action)) {
-						let path = frappe.boot.beam.client[r[0].action][0]
-						resolve(path.split('.').reduce((o, i) => o[i], window)(r)) // calls (first) custom built callback registered in hooks
-					} else {
-						resolve(this[String(r[0].action)](r)) // TODO: this only calls the first function
+			if (isLoginPath) {
+				frappe.xcall('scan_login', { barcode: sCode }).then(r => {
```
Alchez: Similar to the `xcall` below, this can just be the following:

```suggestion
				frappe.xcall('beam.beam.scan.user_login.scan_login', { barcode: sCode }).then(r => {
```

beam/public/js/scan/scan.js
```diff
@@ -3,28 +3,45 @@
...
+
 		observer.observe(document.body, {
 			childList: true,
 			subtree: true,
 		})
+
+		if (isLoginPath) {
+			observer.disconnect()
+			resolve(document.body)
+		}
```
Alchez: If it's the login route, should the observer even be registered then?

beam/beam/scan/user_login.py
```diff
@@ -0,0 +1,34 @@
...
+	user = get_barcode_context(barcode)
+	if not user:
+		frappe.throw("Wrong barcode", title="Login Error")
+
+	if user["doc"].doctype != "User":
+		frappe.throw("Wrong barcode", title="Login Error")
+
+	user_doc = frappe.get_doc("User", user["doc"].name)
+	if not user_doc:
+		frappe.throw("User doesn't exist", title="Login Error")
```
Alchez: `frappe.get_doc` will fail with an error anyway if the record doesn't exist, so this check can be removed

Alchez (changes requested):

beam/beam/overrides/user.py
lauty95: You are right! It's more clean.

beam/hooks.py
lauty95: Got it

lauty95:
IP Restriction added.

To set it up, you need to add the IP in Beam Settings for each company.

If no IP is set, users can log in by scanning without this restriction.

Partial IP is accepted too.

Screenshot 2025-01-29 at 11 34 47 AM

beam/beam/scan/user_login.py
```diff
@@ -0,0 +1,39 @@
...
+	)
+
+	user = get_barcode_context(barcode)
+	if not user:
+		frappe.throw("Wrong barcode", title="Login Error")
+
+	if user["doc"].doctype != "User":
+		frappe.throw("Wrong barcode", title="Login Error")
+
+	company = get_default_company()
```
fproldan: First, check if the user has a related employee and use that employee's company. If not, fall back to the default company.

beam/hooks.py
```diff
@@ -123,7 +123,7 @@
 			"beam.beam.overrides.inventory_dimension.reset_receiving_map",
 		],
 	},
-	("Item", "Warehouse"): {
+	("Item", "Warehouse", "User"): {
 		"validate": ["beam.beam.barcodes.create_beam_barcode"],
```
fproldan, line 127: Do we need to add a check to skip barcode creation for Administrator and Guest? @agritheory

agritheory: Yep, thanks for writing this up!

lauty95: I did some tests reinstalling the site and running `bench execute 'beam.tests.setup.before_test'` and the users "Administrator" and "Guest" have no barcodes.

I think it's due to the hook is created by beam, and beam is installed after frappe does. And the users "Administrator" and "Guest" are created by frappe

lauty95: Added

beam/beam/scan/user_login.py
lauty95: Ready!

agritheory:
@lauty95 One more addition: let's a checkbox to enable/ disable this feature, which should be disabled by default. (so I think an "enabled" field that is 0/false by default)

lauty95:
> @lauty95 One more addition: let's a checkbox to enable/ disable this feature, which should be disabled by default. (so I think an "enabled" field that is 0/false by default)

Added

agritheory:
Awesome! Merged!

PR #250, opened 2025-02-03
fix: don't set FG qty on skip_transfer WO

agritheory:


beam/www/beam/stores/beam.ts
```diff
@@ -242,7 +242,9 @@ export const useBeamStore = defineStore('beam', () => {
 		}
 		// initialize pending stock entry items with zero quantity
 		message.items.map(item => {
-			item.qty = 0
+			if (!item.is_finished_item) {
```
lauty95, line 245: `is_finished_item` is not defined in types

I guess that it should be added in `StockEntryItem` as a number:
https://github.com/agritheory/beam/blob/ee51cfcb17c4a30f20abf86a13c9fe3a1260e198/beam/www/beam/types/frappe.ts#L71

lauty95 (approved):

beam/www/beam/stores/beam.ts
agritheory: @lauty95 I added it as boolean

PR #251, opened 2025-02-11
feat: repack page

lauty95:
# Repack Page

### Mappers:
First, I need to work with `store.cache.mappers`. This is because when you scan an item or a warehouse, the data is stored there. So, I modified this line in `scan.ts`:

```javascript
const documentId = computed(() => {
   const currentRoute = store.router.currentRoute.value
   return currentRoute.params.id || currentRoute.query.id || **store.router.currentRoute.value?.name** || ''
})
```

When you are on the **Move** or **Repack** pages (which do not have an `id`), the route path is used to determine the current page.

With this approach, the data for the Move and Repack pages remain separate.

### Repack
Screenshot 2025-02-11 at 1 55 57 PM

The page has:
- **Item to Repack** - Link - is an async dropdown, so you need to start typing or scan an item to search for it.
- **Qty** - A component from `aform`. I still need to work on its styles.
- **BOM** (Optional) - Link - A dropdown list populated when entering the page. It automatically fills the item list with all the BOM items. (Demo below)
- **Source Warehouse** - Link - Can be selected manually or scanned.
- **Target Warehouse** - Link - Can be selected manually or scanned.
- Items list -ListView - Filled when click the **Add** button

https://github.com/user-attachments/assets/791b0279-7813-4a21-873b-4bff512461ba

### Store
I added the function `getStockEntryItems` to retrieve the items when selecting a BOM. To fetch this data, I call the ERPNext method `get_bom_items`.

The data returned by this method is typed as `BomItem` in `types/frappe.ts`.

### Demo

https://github.com/user-attachments/assets/e86afb1b-61bb-4d50-9de1-17b22b5a0e89

### TODO
I'm working on the BOM feature. There are still some bugs when saving a repack populated via BOM.

### DIFF
--- a/beam/hooks.py
+++ b/beam/hooks.py
@@ -625,7 +625,7 @@ beam_mobile = {
 			"path": "/repack",
 			"name": "repack",
 			"component": "Repack",
-			"meta": {"requiresAuth": True, "doctype": "Stock Entry", "view": "list"},
+			"meta": {"requiresAuth": True, "doctype": "Stock Entry", "view": "form"},
 		},
 		{
 			"path": "/:catchAll(.*)*",
--- a/beam/hooks.py
+++ b/beam/hooks.py
@@ -625,7 +625,7 @@ beam_mobile = {
 			"path": "/repack",
 			"name": "repack",
 			"component": "Repack",
-			"meta": {"requiresAuth": True, "doctype": "Stock Entry", "view": "form"},
+			"meta": {"requiresAuth": True, "doctype": "Stock Entry", "view": "list"},
 		},
 		{
 			"path": "/:catchAll(.*)*",
--- a/beam/www/beam/pages/Repack.vue
+++ b/beam/www/beam/pages/Repack.vue
@@ -12,7 +12,11 @@
 		<div class="container">
 			<template v-if="itemList">
 				<div class="dd-container">
-					<ADropdown label="Item to Repack" :items="itemList" v-model="currentItem.item_code" :isAsync="true"
+					<ADropdown
+						label="Item to Repack"
+						:items="itemList"
+						v-model="currentItem.item_code"
+						:isAsync="true"
 						:filterFunction="loadItems" />
 					<BeamBtn class="clear-button" @click="clearCurrentItem('item_code')"> X </BeamBtn>
 				</div>
@@ -45,12 +49,13 @@
 </template>
 
 <script setup lang="ts">
+import type { ListViewItem } from '@stonecrop/beam'
 import { ref, computed, onMounted, watch } from 'vue'
-import { useBeamStore } from '@/stores/beam'
-import { useBeamToast } from '@/utils/toast.js'
-import { ListViewItem } from '@stonecrop/beam'
+
 import ControlButtons from '@/components/ControlButtons.vue'
+import { useBeamStore } from '@/stores/beam'
 import type { ControlButton, DocActionResponse, StockEntry } from '@/types'
+import { useBeamToast } from '@/utils/toast.js'
 
 const toast = useBeamToast()
 const store = useBeamStore()
@@ -58,8 +63,8 @@ const currentItem = ref({ item_code: '', qty: 0, bom: '' })
 const items = ref([])
 const componentKey = ref(0)
 const stockEntry = computed(
-	(): StockEntry =>
-		(store.cache.mappers['repack'] as StockEntry) || {
+	() =>
+		store.cache.mappers.repack || {
 			name: '',
 			stock_entry_type: 'Repack',
 			items: [],
@@ -73,19 +78,18 @@ const bomList = ref<string[]>([])
 const warehouseList = ref<string[]>([])
 
 onMounted(async () => {
-	store.$patch(state => state.cache.mappers['repack'] = stockEntry.value)
+	store.$patch(state => (state.cache.mappers.repack = stockEntry.value))
 	await loadBOMs()
 	await loadWarehouses()
 })
 
 const loadItems = async (search: string) => {
 	if (!search) return []
-	const itemsData = await store.getAll<{ name: string }[]>('Item', {
+	const items = await store.getAll<{ name: string }[]>('Item', {
 		filters: JSON.stringify([['item_code', 'like', `${search}%`]]),
 	})
-	const itemsMapped = itemsData.map(item => item.name)
-	itemList.value = itemsMapped
-	return itemsMapped.filter(item => item.toLocaleLowerCase().startsWith(search.toLocaleLowerCase()))
+	itemList.value = items.map(item => item.name)
+	return itemList.value.filter(item => item.toLocaleLowerCase().startsWith(search.toLocaleLowerCase()))
 }
 
 const loadBOMs = async () => {
@@ -100,13 +104,12 @@ const loadWarehouses = async () => {
 	warehouseList.value = warehouses.map(warehouse => warehouse.name)
 }
 
-const clearField = (field: 'from_warehouse' | 'to_warehouse') => store.$patch(state => state.cache.mappers.repack[field] = '')
-
-const clearCurrentItem = (field: 'item_code' | 'bom') => currentItem.value[field] = ''
-
-const substractCurrentItem = () => currentItem.value.qty > 0 ? currentItem.value.qty-- : 0
+const clearField = (field: 'from_warehouse' | 'to_warehouse') =>
+	store.$patch(state => (state.cache.mappers.repack[field] = ''))
 
 const addCurrentItem = () => currentItem.value.qty++
+const substractCurrentItem = () => (currentItem.value.qty > 0 ? currentItem.value.qty-- : 0)
+const clearCurrentItem = (field: 'item_code' | 'bom') => (currentItem.value[field] = '')
 
 const create = async () => {
 	const body: StockEntry = {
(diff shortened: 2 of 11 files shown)


lauty95:
Fixed some styles to Quantity input

Screenshot 2025-02-11 at 3 26 48 PM

beam/www/beam/stores/scan.ts
```diff
@@ -19,7 +19,7 @@ export const useScanStore = defineStore('scan', () => {
 
 	const documentId = computed(() => {
 		const currentRoute = store.router.currentRoute.value
-		return currentRoute.params.id || currentRoute.query.id || ''
+		return currentRoute.params.id || currentRoute.query.id || store.router.currentRoute.value?.name || ''
```
Alchez: Would the following also work:

```suggestion
		return currentRoute.params.id || currentRoute.query.id || currentRoute.name || ''
```

beam/hooks.py
```diff
@@ -625,7 +625,7 @@
 			"path": "/repack",
 			"name": "repack",
 			"component": "Repack",
-			"meta": {"requiresAuth": True, "doctype": "Stock Entry", "view": "list"},
+			"meta": {"requiresAuth": True, "doctype": "Stock Entry", "view": "form"},
```
Alchez, line 628: @lauty95 I had to change the setup here, since trying to scan an item was somehow leading to the `filter` action. Not sure how it was working for you locally though :thinking:

beam/www/beam/types/scan.ts
```diff
@@ -53,7 +53,7 @@ export type ScanContext = {
 }
 
 export type ScanConfig = {
-	client?: Record<string, string[]>[]
+	client?: Record<string, string[]> | never[]
```
Alchez, line 56: Note: if no custom `beam_client` hooks exist, the response from the server is always an empty list. So accounting types for that.

beam/www/beam/pages/Repack.vue
```diff
@@ -8,24 +8,293 @@
...
+
+const clearItem = () => {
+	currentItem.value.bom = ''
+	items.value = []
+	store.$patch(state => (state.cache.mappers.repack.items = []))
+}
+
+const controlButtons = computed((): ControlButton[] => {
+	const buttons = [
+		{ label: 'CLEAN', color: { background: '#4791FF', text: 'var(--sc-btn-color)' }, action: clearItem },
```
Alchez: I don't think this button should show up until an item has been added, otherwise it might get confusing

beam/www/beam/pages/Repack.vue
```diff
@@ -8,24 +8,293 @@
...
+					<ANumericInput label="Quantity" v-model="currentItem.qty" />
+					<BeamBtn class="clear-button" @click="addCurrentItem"> + </BeamBtn>
+				</div>
+				<div class="dd-container">
+					<ADropdown label="BOM (Optional)" :items="bomList" v-model="currentItem.bom" />
+					<BeamBtn class="clear-button" @click="clearCurrentItem('bom')"> X </BeamBtn>
+				</div>
+			</template>
+			<div class="dd-container">
+				<ADropdown label="Source Warehouse" :items="warehouseList" v-model="stockEntry.from_warehouse" />
```
Alchez, line 30: Should the selection map to the parent document, or to the individual items? This might only be a concern for repacks with multiple input/output items that may exist in different warehouses.

beam/www/beam/pages/Repack.vue
```diff
@@ -8,24 +8,292 @@
...
 
-const route = useRoute()
-const stockEntryId = route.params.id
-const items = ref<ListViewItem[]>([])
-const stockEntry = ref<Partial<StockEntry>>({})
+onMounted(async () => {
+	store.$patch(state => (state.cache.mappers.repack = stockEntry.value))
+	await loadBOMs()
+	await loadWarehouses()
+})
```
Alchez: This might be better run before mount maybe to allow enough time for the API calls. But if it that doesn't work, then this way is fine too.

Alchez (changes requested):

agritheory:
Manual testing, scanning an item and warehouse work as expected. Scanning a Handling Unit raised an error. 


```
  File "apps/frappe/frappe/app.py", line 110, in application
    frappe.handler.handle()
  File "apps/frappe/frappe/handler.py", line 50, in handle
    data = execute_cmd(cmd)
  File "apps/frappe/frappe/handler.py", line 86, in execute_cmd
    return frappe.call(method, **frappe.form_dict)
  File "apps/frappe/frappe/__init__.py", line 1726, in call
    return fn(*args, **newargs)
  File "apps/frappe/frappe/utils/typing_validations.py", line 31, in wrapper
    return func(*args, **kwargs)
  File "apps/beam/beam/beam/scan/__init__.py", line 29, in scan
    return get_form_action(barcode_doc, context_dict)  # TODO: add current_qty argument here
  File "apps/beam/beam/beam/scan/__init__.py", line 177, in get_form_action
    target = get_stock_entry_item_details(context.doc, hu_details.item_code)
  File "apps/beam/beam/beam/scan/__init__.py", line 115, in get_stock_entry_item_details
    stock_entry = StockEntry(frappe._dict(doc))
TypeError: 'NoneType' object is not iterable
```

beam/www/beam/pages/Repack.vue
agritheory: Parent document. We can only get so sophisticated in a mobile view.

beam/www/beam/pages/Repack.vue
agritheory: What event would run before `onMounted`?

Alchez: Vue's lifecycle hooks offer an [`onBeforeMount`](https://vuejs.org/api/composition-api-lifecycle.html#onbeforemount) event that runs before the DOM elements are rendered. I was thinking of adding the prefetch stuff in here, but also this event isn't SSR-friendly if we wanted to consider it for something later down the line.

beam/www/beam/pages/Repack.vue
Alchez: @agritheory @lauty95 Then I think we should validate/confirm when a different warehouse is detected, either through scan or selection maybe?

beam/hooks.py
lauty95: I used "form" instead of "list" because with "list" I get `filter` action instead of `add_or_increment`, which I was ussing to set the items.

Though, I assumed that I should use form instead of list because `/move` path is using `form`.

I uploaded a little demo.

https://drive.google.com/file/d/1r6VAAAAIim40I4M7yvYG4_UxZhPsIoYf/view?usp=sharing

beam/www/beam/pages/Repack.vue
lauty95: What about filling the list of warehouses in the store here:

https://github.com/agritheory/beam/blob/9cd2df94f307e38d894fb6dd88014f610a5687e8/beam/www/beam/stores/init.ts#L13-L20

We could use this data in both Repack.vue and Move.vue:

https://github.com/agritheory/beam/blob/9cd2df94f307e38d894fb6dd88014f610a5687e8/beam/www/beam/pages/Move.vue#L64-L69

This way, we avoid making extra API calls.

Another approach could be using an isLoading state, setting it before and after the API call, and showing a "Loading" label or a spinner instead of the inputs interface.

Alchez: I like the idea of setting up the warehouses during init, if the list of warehouses wouldn't be expected to change often. I think it _should_ be available before the loading state is reached, but we can run some tests for if that is required.

lauty95:
> Manual testing, scanning an item and warehouse work as expected. Scanning a Handling Unit raised an error.
> 
> ```
> Traceback (most recent call last):
>   File "apps/frappe/frappe/app.py", line 110, in application
>     frappe.handler.handle()
>   File "apps/frappe/frappe/handler.py", line 50, in handle
>     data = execute_cmd(cmd)
>   File "apps/frappe/frappe/handler.py", line 86, in execute_cmd
>     return frappe.call(method, **frappe.form_dict)
>   File "apps/frappe/frappe/__init__.py", line 1726, in call
>     return fn(*args, **newargs)
>   File "apps/frappe/frappe/utils/typing_validations.py", line 31, in wrapper
>     return func(*args, **kwargs)
>   File "apps/beam/beam/beam/scan/__init__.py", line 29, in scan
>     return get_form_action(barcode_doc, context_dict)  # TODO: add current_qty argument here
>   File "apps/beam/beam/beam/scan/__init__.py", line 177, in get_form_action
>     target = get_stock_entry_item_details(context.doc, hu_details.item_code)
>   File "apps/beam/beam/beam/scan/__init__.py", line 115, in get_stock_entry_item_details
>     stock_entry = StockEntry(frappe._dict(doc))
> TypeError: 'NoneType' object is not iterable
> ```

Added this feature.

Could you test it and tell me if is ok this way?

beam/www/beam/pages/Repack.vue
lauty95: @Alchez 

Added on init:
https://github.com/agritheory/beam/blob/e7aff00b1e6a88603d103b1e52eed886b7a9b1cb/beam/www/beam/stores/beam.ts#L117-L121

I thought about fetching those fields just in case we need them in the frontend.

So, in Repack.vue and Move.vue, it ended up like this:
https://github.com/agritheory/beam/blob/e7aff00b1e6a88603d103b1e52eed886b7a9b1cb/beam/www/beam/pages/Repack.vue#L76-L80

Filtering by `is_group`

However, I could keep the fetching filtered by is_group and save store.warehouseList as a string array.

Alchez: I think it might be better the way you've implemented it, since getting all the data once and re-using that in different contexts is better than realising another query is required somewhere downstream in the workflow.

PR #260, opened 2025-02-20
feat: add filters for beam list views

### LINKED ISSUE #210: [mobile] Filtering on various listviews
### Manufacture
- Via UI:
  - [x] Status
  - [x] All/Past/Today/Future (expected start date)
  - [x] Assigned
- Via scanning:
  - [ ] Workstation
  - [ ] Production Item
  - [ ] Source Item Warehouse

### Demand
- Via UI:
  - [x] Status
  - [x] All/Past/Today/Future (expected start date)
  - [x] Assigned
- Via scanning:
  - [ ] Workstation
  - [ ] Production Item
  - [ ] Source Item Warehouse

### Receive
- Via UI:
  - [x] Status
  - [x] All/Past/Today/Future (expected delivery date)
  - [x] Assigned
- Via scanning:
  - [ ] Workstation
  - [ ] Production Item
  - [ ] Source Item Warehouse

### Ship
- Via UI:
  - [x] Status
  - [x] All/Past/Today/Future (expected delivery date)
  - [x] Assigned
- Via scanning:
  - [ ] Workstation
  - [ ] Production Item
  - [ ] Source Item Warehouse

Alchez:
Closes #210.

### DIFF
--- a/beam/beam/scan/config.py
+++ b/beam/beam/scan/config.py
@@ -34,10 +34,8 @@ def get_scan_doctypes():
 			scannable_doctypes.add(key)
 			[frm_doctypes.add(value) for value in values.keys()]
 
-	_scan_last = frappe.get_all(
-		"BEAM Settings",
-		["show_scan_output"],
-	)
+	# TODO: should this be filtered against a specific company?
+	_scan_last = frappe.get_all("BEAM Settings", fields=["show_scan_output"])
 	scan_last = _scan_last[0] if _scan_last else {"show_scan_output": False}
 
 	return {
--- a/beam/beam/scan/config.py
+++ b/beam/beam/scan/config.py
@@ -34,9 +34,16 @@ def get_scan_doctypes():
 			scannable_doctypes.add(key)
 			[frm_doctypes.add(value) for value in values.keys()]
 
+	_scan_last = frappe.get_all(
+		"BEAM Settings",
+		["show_scan_output"],
+	)
+	scan_last = _scan_last[0] if _scan_last else {"show_scan_output": False}
+
 	return {
 		"scannable_doctypes": list(scannable_doctypes),
 		"listview": list(listview_doctypes),
 		"frm": list(frm_doctypes),
 		"client": beam_client,
+		**scan_last,
 	}
--- a/beam/beam/demand/demand.py
+++ b/beam/beam/demand/demand.py
@@ -694,7 +694,6 @@ def get_demand(*args, **kwargs) -> list[Demand]:
 		page = int(kwargs.get("page", 1))
 	except ValueError:
 		page = 1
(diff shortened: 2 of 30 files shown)


agritheory:
image

An excellent start. I think we can reduce the white space and switch to a two column layout. Another default filter that we had no included in the spec issue is "assigned" / "assigned to me".

Alchez:
@agritheory A few updates:

- I've added a user filter for the `Manufacture`, `Demand`, `Receive` and `Ship` views.
  - It was straightforward for the Manufacture list view since that would use Frappe's resource API and the assignment data structure is documented. But I wasn't sure how assignments are stored in the demand DB.
  - I see an `assigned` string field inside the demand tables, but I'm not sure if that's supposed to represent a single user or a list of users (comma-separated?)
- I've also added a PR for making the filters a two-column layout over at https://github.com/agritheory/stonecrop/pull/243.
- I've added some design questions for scanning filters in the [connected issue](https://github.com/agritheory/beam/issues/210#issuecomment-2674255848)

agritheory:
z-index issue: 

image

agritheory:
@Alchez 1) I think we need to z-index the container, this dropdown shouldn't be pushing the list down.
2) I think the break logic needs to be sync'd with the Stonecrop PR, where it's at 479px.

@crabinak FYI

image

Alchez:
@agritheory @crabinak I added a fix for the dropdown pushing the list content over at https://github.com/agritheory/stonecrop/pull/250.

beam/beam/scan/config.py
```diff
@@ -34,10 +34,8 @@ def get_scan_doctypes():
 			scannable_doctypes.add(key)
 			[frm_doctypes.add(value) for value in values.keys()]
 
-	_scan_last = frappe.get_all(
-		"BEAM Settings",
-		["show_scan_output"],
-	)
+	# TODO: should this be filtered against a specific company?
```
agritheory, line 37: Associating an arbitrary user to company isn't easy, it requires an Employee record. I think adding a comment on the field is better that it will be visible in all companies.

agritheory:
@Alchez Not sure what happened in your cleanup, but the scan input component is now broken. These should be showing doctype and docname.

image

Alchez:
@agritheory Hmm, I just looked at the code again and it seems like you're only returning those values (doctype: docname) for list action responses, whereas the `Manufacture` page is setup as a form page, which is why it doesn't get the `parenttype` and `parent` values from the server correctly.

agritheory:
@Alchez After updating the stonecrop packages, I am still seeing three filters on the same row but the z-index issue is fixed

image

Alchez:
@agritheory Oh sorry, I didn't know we wanted to restrict the number of filters in each row as well. Should there be two filters per row maximum?

PR #263, opened 2025-02-27
[app] scan items in stock reconciliation form

lauty95:
# Stock Reconciliation Handling Scanning

## Demo

https://github.com/user-attachments/assets/8dfc1387-504f-49c4-b676-0aff5bac7311


agritheory:
@lauty95 Questions about your demo: It looks like scanning at item clears the table, after you've scanned the WH. I don't think that's right, it should add_or_increment instead. Have you tested this with a handling unit?

lauty95:
@agritheory 
1 - Yes! It clears the table. I'll update it.
2 - Mmm I haven't tested this with a handling unit. I'll check it out

beam/public/js/scan/scan.js
```diff
@@ -208,6 +208,24 @@ class ScanHandler {
 				frappe.model.set_value(row.doctype, row.name, 's_warehouse', barcode_context.target)
 				frappe.model.set_value(row.doctype, row.name, 't_warehouse', barcode_context.target)
 			}
+		} else if (barcode_context.doctype == 'Stock Reconciliation Item') {
+			frappe.model.set_value(cur_frm.doc.doctype, cur_frm.doc.name, 'set_warehouse', barcode_context.context.doc.name)
```
Alchez: If you're setting a value in the parent document, then you can re-use the `cur_frm` APIs:

```suggestion
			cur_frm.set_value('set_warehouse', barcode_context.context.doc.name)
```

Alchez (changes requested):

lauty95:
@agritheory 

Demo of changes:

https://github.com/user-attachments/assets/4d957cc0-3143-41b9-925d-580ab7786c30

It would be great if you could test it.

One thing to note: When you scan an HU, it triggers `add_or_associate` action, which doesn't increment qty , but if you scan a WH that contains that item, it does increment the qty.

@Alchez I'm not using `cur_frm.set_value` because that method calls functions from `stock_reconciliation.js` that reset the table.

Alchez:
> @Alchez I'm not using cur_frm.set_value because that method calls functions from stock_reconciliation.js that reset the table.

Oh that's weird. In that case, do add a comment around that code that specifies this, in case we need to figure out why it's happening in a non-standard way.

agritheory:
I am experiencing #268 on this branch while attempting to review it.

Alchez (approved):

Alchez:
@agritheory @lauty95 I'm unable to replicate #268 on this branch, but I did see that error on #264. But otherwise, this PR should be good to proceed.

PR #271, opened 2025-05-30
254 renovate stock reconciliation form view workflows

### LINKED ISSUE #254: Renovate Stock Reconciliation form view workflows
- Scanning a warehouse should populate the warehouse field and load the items and handling units.
- Scanning an item does not load its valuation rate (what `get_item_details` endpoint needs to be called?)

lauty95:
Resolve #254 

Note on `valuation rate`:
I used the get_valuation_rate function from ERPNext, which fetches the valuation rate from the Bin doctype using the item's default warehouse. With Beam's demo data, the only item in the Bin doctype that has a valuation rate available for testing is "Ice Water".

# Demo

https://github.com/user-attachments/assets/b8c4bee7-fc3c-4ae2-98c8-988771f85d96

### DIFF
--- a/beam/beam/scan/__init__.py
+++ b/beam/beam/scan/__init__.py
@@ -217,6 +217,14 @@ def get_form_action(barcode_doc: frappe._dict, context: frappe._dict) -> list[di
 				}
 			)
 		target.barcode = barcode_doc.barcode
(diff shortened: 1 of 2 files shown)


agritheory (approved):
Looks great!

PR #280, opened 2025-07-24
Add completed_qty to test fixtures 

lauty95:
ERPNext has updated recently: https://github.com/frappe/erpnext/pull/48745

If you try to execute bench execute 'beam.tests.setup.before_test' with ERPNext updated to v15.71.1, it will throw an error:

Screenshot 2025-07-24 at 1 40 00 PM

`erpnext.manufacturing.doctype.job_card.job_card.OperationSequenceError: Job Card PO-JOB00002: As per the sequence of the operations in the work order MFG-WO-2025-00001, complete the operation Gather Pie Filling Ingredients before the operation Cook Pie Filling Operation.`

This error occurs because Work Order has a new function,`validate_operations_sequence`:

https://github.com/frappe/erpnext/blob/3825490f0dbbd5530b739e928d012f639b6efafe/erpnext/manufacturing/doctype/work_order/work_order.py#L173

This function assigns`sequence_id` to all operations that don't have one assigned, and then we have this function:

https://github.com/frappe/erpnext/blob/ad45063c76880a443c772a0ea3aa7ab6ca33da3c/erpnext/manufacturing/doctype/job_card/job_card.py#L1032

That function was created one year ago, but it wasn't a problem because it has this condition:

```js
    if not (self.work_order and self.sequence_id):
        return
```

Since we didn't assign`sequence_id` to our operations, this validation wasn't a problem.

But this function has a condition before:

```js
	def validate_sequence_id(self):
		if self.is_corrective_job_card:
			return
```

That is the reason for my change.

### DIFF
--- a/beam/beam/overrides/inventory_dimension.py
+++ b/beam/beam/overrides/inventory_dimension.py
@@ -1,3 +1,6 @@
+# Copyright (c) 2024, AgriTheory and contributors
+# For license information, please see license.txt
+
 from beam.beam.demand.demand import build_demand_allocation_map
 from beam.beam.demand.receiving import reset_build_receiving_map
 
--- a/beam/beam/scan/user_login.py
+++ b/beam/beam/scan/user_login.py
@@ -1,3 +1,6 @@
+# Copyright (c) 2024, AgriTheory and contributors
+# For license information, please see license.txt
+
 import frappe
 from erpnext import get_default_company
 from frappe.core.doctype.user.user import get_restricted_ip_list
--- a/beam/tests/setup.py
+++ b/beam/tests/setup.py
@@ -702,14 +702,13 @@ def create_production_plan(settings, prod_plan_from_doc):
 			job_card.append(
 				"time_logs",
 				{
-					"completed_qty": wo.qty,
+					#"completed_qty": wo.qty,
 					"from_time": start_time,
 					"to_time": start_time + datetime.timedelta(minutes=time_in_mins),
 					"time_in_mins": time_in_mins,
 					"remaining_time_in_mins": time_in_mins,
 				},
 			)
-			job_card.is_corrective_job_card = 1
 			job_card.save()
 			start_time = job_card.time_logs[0].to_time + datetime.timedelta(minutes=2)
 			# job_card.submit() # TODO: don't submit for demand tests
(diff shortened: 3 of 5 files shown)


agritheory:
An alternative is to add sequence id in the fixturing: https://github.com/agritheory/beam/blob/version-14/beam/tests/setup.py#L343

lauty95:
@agritheory I tested the sorting with `order_by`:

`job_cards = frappe.get_all("Job Card", {"work_order": wo.name}, order_by="sequence_id ASC, creation ASC")`

It didn’t solve the issue, so I kept debugging.

The problem comes from `validate_sequence_id` in  [job_card.py:L1058](https://github.com/frappe/erpnext/blob/d4519e5d3dc789e49093514edfc57c663ecd5ef1/erpnext/manufacturing/doctype/job_card/job_card.py#L1058):

```py
row.status != "Completed"  # This condition is true, it is not completed yet
row.completed_qty < current_operation_qty  # This is also true, and I believe this is what should be avoided
```

Case where it fails:

- row.completed_qty = 0
- current_operation_qty = 160.0
- Happens on `Job Card PO-JOB00002` from Work Order `MFG-WO-2025-00001`.

`current_operation_qty` comes from `self.total_completed_qty`, which is set in `validate_time_logs` ([L232](https://github.com/frappe/erpnext/blob/d4519e5d3dc789e49093514edfc57c663ecd5ef1/erpnext/manufacturing/doctype/job_card/job_card.py#L232)):

```py
if d.completed_qty and not self.sub_operations:
	self.total_completed_qty += d.completed_qty
```

This causes the validation to fail.

**Fix applied:** when adding `time_logs` to the job card, do not set "completed_qty":

```py
job_card.append(
"time_logs",
{
	#"completed_qty": wo.qty,
	"from_time": start_time,
	"to_time": start_time + datetime.timedelta(minutes=time_in_mins),
	"time_in_mins": time_in_mins,
	"remaining_time_in_mins": time_in_mins,
},
)
```

Tests still pass, and Playwright tests run without issues.

Not entirely sure if setting "completed_qty" here is ever required.

agritheory (approved):
I think the completed_qty fix is the correct approach

agritheory:
@lauty95 Please also cherry pick this for a PR against the version-15 branch

PR #281, opened 2025-07-28
fix: controls deleted

### LINKED ISSUE #279: Reroute is not sensitive to "BEAM Mobile User" on login
We don't want to check for anything else. Firefox shows up as a false positive for Android when using a desktop version of FF, which is a deal breaker.

lauty95:


lauty95:
Pytest CI/tests will fail because ERPNext has been updated.

Fixed on #280

agritheory (approved):

PR #286, opened 2025-08-11
Fix job card fixtures

agritheory:


agritheory:
@lauty95 Is this a duplicate? Feel free to incorporate it into another branch if appropriate

Issue #288, opened 2025-08-13
Tests doesn't check as failed when it should


lauty95:
# Example

<img width="1279" height="829" alt="Image" src="https://github.com/user-attachments/assets/2df55e5f-abed-4f50-88e2-a2a0f8ab1897" />

https://github.com/agritheory/beam/actions/runs/16938306186/job/48000277134?pr=287

PR #289, opened 2025-08-21
feat: pipefail

lauty95:
Resolves https://github.com/agritheory/beam/issues/288

### DIFF
--- a/beam/beam/overrides/stock_entry.py
+++ b/beam/beam/overrides/stock_entry.py
@@ -26,11 +26,16 @@ class BEAMStockEntry(StockEntry):
 				if hasattr(sle, 'voucher_detail_no') and sle.voucher_detail_no:
 					for item in self.get("items"):
 						if item.name == sle.voucher_detail_no:
-							if item.handling_unit:
+							# For transfers with both handling_unit and to_handling_unit
+							if item.handling_unit and item.to_handling_unit:
+								if sle.get('warehouse') == item.s_warehouse:
+									# Source warehouse uses original handling_unit
+									sle.handling_unit = item.handling_unit
+								elif sle.get('warehouse') == item.t_warehouse:
+									# Target warehouse uses to_handling_unit
+									sle.handling_unit = item.to_handling_unit
+							elif item.handling_unit:
 								sle.handling_unit = item.handling_unit
-							elif item.to_handling_unit and sle.get('warehouse') == item.t_warehouse:
-								# For transfers, use to_handling_unit for target warehouse
-								sle.handling_unit = item.to_handling_unit
 							break
 		
 		if self.docstatus == 2:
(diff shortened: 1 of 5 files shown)


agritheory:
@lauty95 Can you explain what this is and why it's needed?

lauty95:
> @lauty95 Can you explain what this is and why it's needed?

I'm still testing this, but the issue was in this line:

`pytest --cov=beam --cov-report=xml --disable-warnings -s | tee pytest-coverage.txt`

The | (pipe) was catching the error and writing it to a txt file, so theoretically the process was not failing at the end.

It seems that it's working :) because now it's failing

lauty95:
@agritheory If you want, I could fix the tests in this PR, or if you prefer I can open another PR

agritheory:
@lauty95 Go for it, let us know if you need a hand.

lauty95:
@agritheory 
# Changes

## Easiest changes:

- **Current year** 2024 was hardcoded as a string, changed to current_year variable to update year-by-year automatically.

- **Job Card** Added the completed_qty before submit to meet ERPNext validation requirements.

- `build_demand_allocation_map` was used to "reset" the demand SQLite DB before some tests that needed synchronized data.

## Handling Units:

I added a feature in Stock Entry that correctly saves handling units to Stock Ledger Entries. This was necessary for this two scenarios:

1. `test_stock_entry_material_transfer` (Handling Unit Tests):
Material Transfer operations weren't correctly assigning handling units to Stock Ledger Entries, both source and target warehouse entries used the same handling_unit instead of differentiating between `handling_unit` (source) and `to_handling_unit` (target)

2. **Stock quantities were calculated incorrectly**
The `get_handling_unit()`  function couldn't find expected handling units because they weren't saved correctly in Stock Ledger Entries.

I based these changes in `stock_entry.py` file on tests functions. Please let me know if I should make any other change on test functions or in `stock_entry.py` file.

beam/beam/overrides/stock_entry.py
```diff
@@ -19,6 +19,25 @@ def update_stock_ledger(self):
 		finished_item_row = self.get_finished_item_row()
 		self.get_sle_for_source_warehouse(sl_entries, finished_item_row)
 		self.get_sle_for_target_warehouse(sl_entries, finished_item_row)
+
+		# Add handling_unit to Stock Ledger Entries
```
agritheory, line 23: @lauty95 I want to understand why this is required - my thought is that this _should_ be automatically handled by the Inventory Dimension.

lauty95: @agritheory  

Do you prefer that I remove the changes I made in stock_entry.py?  
That way, we can see the errors in the tests and analyze how to handle them.

agritheory: I read this again, it's my fault for misunderstanding. False alarm. All good.

PR #291, opened 2025-08-31
Update barcodes.py for nested item group handling unit generation

diniemasjuki:
Fix attempt on stock item handling unit barcode generation not working automatically on nested item group of products during item creation. 
Attempt to enable code128 generation for all stock item group seems successful


agritheory:
@diniemasjuki You are making a PR from a version-15-based branch to version-14. This change has already been added in version-15

PR #309, opened 2026-02-04
feat: make handling units respect config, add defaults

agritheory:


beam/beam/boot.py
```diff
@@ -1,9 +1,31 @@
...
+	for setting in beam_settings:
+		settings[setting.company] = {
+			"enable_handling_units": setting.enable_handling_units,
+		}
+	return settings
+
+
+def get_handling_unit_default_print_format():
+	"""Get the default print format for Handling Unit doctype."""
+	return frappe.db.get_value("DocType", "Handling Unit", "default_print_format")
```
lauty95: This field doesn't exist

agritheory: That's supposed to be calling get_meta to get the default print format, yeah.

agritheory: @lauty95 updated

lauty95 (approved):

PR #326, opened 2026-02-23
Mobile repack test

IshwaryaM1030:


lauty95:
Hey @IshwaryaM1030! I found two issues while testing locally:

- The decorator @pytest.mark.order() in test_repack.py isn’t preserving the expected order relative to the earlier tests. Locally, I adjusted it to start from 8, and with that change the previous tests continue working fine.

Screenshot 2026-02-25 at 9 51 28 AM

- Once the order was fixed, I noticed the next error:

```sh
erpnext.stock.doctype.stock_entry.stock_entry.FinishedGoodError: There must be at least 1 Finished Good in this Stock Entry
```

Screenshot 2026-02-25 at 9 49 35 AM

Screenshot 2026-02-25 at 9 33 57 AM


This happens because there isn’t enough test data for the repack scenarios. We should update `create_test_data` to include the proper data for repack tests. That would fix both `test_repack_items_manually` and `test_repack_using_bom`.

Screenshot 2026-02-25 at 9 50 56 AM

Could you make these changes?

beam/tests/mobile/test_repack.py
```diff
@@ -0,0 +1,256 @@
...
+			limit=1,
+		)[0]
+
+	with page.expect_request(
+		lambda request: request.headers.get("x-frappe-cmd") == "beam.beam.scan.scan"
+	):
+		page.evaluate("barcode => scanner.simulate(window, barcode)", barcode)
+	page.wait_for_timeout(800)
+
+	page.get_by_role("button", name="+").click()
```
lauty95, line 52: we’re adding one more item manually. It would be good to assert the quantity label before and after each step.

For example:
- Before scanning, it should show `0/`
- After simulating the scan, it should show `1/`
- After clicking the "+" button, it should show `2/`
- After clicking the "-" button, it should show `1/`

PR #340, opened 2026-04-17
feat inv dim propagation

### LINKED ISSUE #135: Preserve handling unit PK when it is moved in its entirety
Would it be useful to include settings that enable
1. Preserving HU numbers for transfers of complete qty's? Given BEAM generates new numbers for every transaction, not just receipts, repacks, or mfr's
2. Handling Unit number naming series options

robproject:
### Description
* Add inventory dimensions to `get_handling_unit` method, using cached list of dimensions, for update of all target child table documents.
* Refactor Handling Unit inventory dimension creation into method to be shared by any app for Inventory Dimension creation.
* Add `custom_carry_forward` field to `Inventory Dimension` DocType for use in `propagate_inventory_dimensions`.
* Add `propagate_inventory_dimensions` method as `before_submit` hook on `Stock Entry` to enable below behavior. BEAM's `handling_unit.py` is updated to not overwrite `target_handling_unit` if it has a value set already, unless the HU qty and transfer qty don't match.

| custom_carry_forward | Transfer qty | to_handling_unit result |
| -----------------------|--------------|------------------------|
| On                                |   qty == stock (full) |    Same HU (conserved) |
| On                                 | qty != stock (split) | New HU (BEAM overrides) |
| Off                              |     any                       | New HU (BEAM default) |

* Update tests.
* Syntax changes in scan/__init__.py for v16: pass frappe._dict(), use SUM dict in frappe.get_all()
* Similar to https://github.com/agritheory/beam/pull/137 but applies to any configured Inventory Dimension and accounts for qty splitting.


agritheory:
@robproject This is a nice feature!

robproject:
I have two other versions of BEAM that are customized a little differently. Need to try combining them all for this. It started with only this PR but each system diverged.

robproject:
fix(print_format): stop emitting None ahead of ^XA in the Handling Unit 6x4 ZPL format

Label.add() mutates the label and returns None. The format invoked it inside output tags — {{ label.add(...) }} — so Jinja rendered those return values into the raw stream. Together with a newline leaking from the leading {% set %} tag, every generated label began:

\nNoneNone\nNone\nNone\nNone\nNone\n\n^XA...

Zebra firmware discards a format when any data precedes ^XA, and the resulting failure is silent and expensive to diagnose: the CUPS job is accepted and reported completed, the printer never moves, and nothing is logged to /var/log/cups/error_log. Confirmed on a GK420d over USB against a raw queue — capturing the spool file and stripping only the None prefix made the identical label print correctly first time.

Changes:

All six add() calls converted to {%- set _ = label.add(...) -%}, so the side effect happens without rendering the return value.
Whitespace control tightened on the two leading set tags so the stream starts exactly at ^XA.

PR #343, opened 2026-06-22
Add printer wizard

agritheory:


agritheory:
Requires https://github.com/agritheory/test_utils/pull/211

lauty95:
https://github.com/agritheory/beam/blob/c82ed55db984e9dda869975e31275a96b6240312/cups/cups/Containerfile#L42

I tried spinning up the CUPS container to test this PR. The Dockerfile hasn't been updated in 2 years and has some deprecated packages that can be safely ignored for this PR:

- Line 42 references a nonexistent file `avahi-daemon.conf`. I commented it out.
- `printer-driver-all` and `printer-driver-gutenprint` are unavailable for Debian arm64. I commented those out as well.

Screenshot 2026-06-23 at 10 26 01 AM

agritheory:
@lauty95 Ah. I'm going to need to incorporate that into the tests, I was testing locally with a local service, not a container.

beam/public/js/network_printer_settings_custom.js
```diff
@@ -9,12 +9,337 @@ function set_location_from_cache(frm, printer_name) {
...
+	if (!status) {
+		return
+	}
+
+	const color = status.indicator_color || 'green'
+	const label = frappe.utils.escape_html(status.indicator_label || __('Unknown'))
+	const make_model = frappe.utils.escape_html(status.make_model || '')
+	const description = frappe.utils.escape_html(status.description || '')
+	const accepting = status.is_accepting_jobs ? __('Accepting jobs') : __('Rejecting jobs')
+	const reasons = status.state_reasons_display ? frappe.utils.escape_html(status.state_reasons_display) : ''
```
lauty95:
```suggestion
	const reasons = (status.state_reasons_display && status.state_reasons_display !== 'none')
		? frappe.utils.escape_html(status.state_reasons_display)
		: ''
```

beam/public/js/network_printer_settings_custom.js
```diff
@@ -9,12 +9,337 @@ function set_location_from_cache(frm, printer_name) {
...
+		lines.push(reasons)
+	}
+
+	const html = `<div class="text-muted small">${lines.filter(Boolean).join('<br>')}</div>`
+
+	if (!frm.cups_status_wrapper) {
+		frm.cups_status_wrapper = $(`<div class="cups-status-dashboard"></div>`).insertAfter(
+			frm.layout.wrapper.find('.form-layout').first()
+		)
+	}
```
lauty95:
```suggestion
	if (!frm.cups_status_wrapper || !$.contains(document, frm.cups_status_wrapper[0])) {
		frm.cups_status_wrapper = $(`<div class="cups-status-dashboard" style="padding: 8px 15px; border-top: 1px solid var(--border-color);"></div>`)
			.appendTo(frm.layout.wrapper.find('.form-page'))
	}
```

lauty95: Image

Status panel not appearing on the form

beam/beam/overrides/network_printer_settings.py
```diff
@@ -1,36 +1,863 @@
...
+			return parsed.hostname
+	except Exception:
+		pass
+	match = re.search(r"://([^:/]+)", device_uri)
+	if not match:
+		return None
+	host = match.group(1)
+	if host.endswith(".local"):
+		return host
+	return host
```
lauty95, line 146:
```suggestion
	host = match.group(1)
	return host
```

beam/beam/printer_queue.py
```diff
@@ -0,0 +1,458 @@
...
+
+@frappe.whitelist()
+def cancel_printer_job(server_ip, port, job_id):
+	check_printer_queue_write_permission()
+	conn = cups_connection(server_ip, int(port))
+	cups = require_cups()
+	try:
+		conn.cancelJob(int(job_id))
+	except cups.IPPError as exc:
+		frappe.throw(_("Could not cancel print job: {0}").format(exc))
```
lauty95:
```suggestion
```

beam/beam/overrides/network_printer_settings.py
```diff
@@ -1,36 +1,863 @@
...
+				"printer_type": printer_type,
+				"device_uri": device_uri,
+			}
+		)
+		doc.insert()
+	except Exception as exc:
+		try:
+			conn.deletePrinter(printer_name)
+		except Exception:
+			pass
```
lauty95:
```suggestion
			frappe.log_error(
        title="CUPS rollback failed",
        message=f"Could not delete orphaned queue {printer_name} after NPS insert failure",
    )
```

lauty95 (changes requested):

agritheory:
@lauty95 Requested changes added, CUPS tests use `testcontainers` to run the containerfile that's part of the repo which is also available in the GHCR after this branch is merged.

lauty95:
@agritheory  I saw the tests got stuck/hung — maybe we could add `pytest-timeout` as a dependency, so we can use the `--timeout` flag in pytest and prevent it from hanging until we happen to notice.

We could also add `timeout-minutes` to the tests job — considering it spins up the whole bench plus its configuration, a 30-minute limit would be fine.

Minor detail: `concurrency` is commented out in `pytest.yaml`, we could uncomment it to avoid parallel workflows when there are several pushes.

This is just to prevent the hang and stop it from running for 2h. We know the real hang starts after `test_printer_cups_integration.py`, probably in `test_printer_queue.py`, but I haven't tested it yet.

agritheory:
@lauty95 Tests pass!

lauty95 (approved):

PR #344, opened 2026-06-24
feat: port scan to login from mobile_v15 to version-15

agritheory:


beam/tests/test_user_login.py
```diff
@@ -0,0 +1,154 @@
...
+from beam.beam.barcodes import create_beam_barcode
+from beam.install import create_beam_mobile_user_role
+
+
+TEST_USER_EMAIL = "scan_login_test@example.com"
+
+
+@pytest.fixture()
+def beam_settings():
+	company = frappe.defaults.get_defaults().get("company")
```
lauty95:
```suggestion
from frappe.auth import CookieManager

from erpnext import get_default_company
from beam.beam.barcodes import create_beam_barcode
from beam.install import create_beam_mobile_user_role


TEST_USER_EMAIL = "scan_login_test@example.com"


@pytest.fixture()
def beam_settings():
	company = get_default_company()
```

lauty95 (commented):
@agritheory , I tested it locally and it works:

I tested from settings: allow all users, beam users, and everyone. In all cases the validations triggered correctly from the login.

Image

I tested IP blocking, works ok.

Image

I tried setting auto_barcode_doctypes = '["Item", "Warehouse"]' and then your change to auto_barcode_doctypes = '["Item", "Warehouse", "User"]'. The problem is that if a site already has users loaded that don't have a barcode, applying this change won't give them a barcode.
Might need a patch to backfill barcodes for existing users on migrate.

Image

beam/tests/test_user_login.py
agritheory: @lauty95 Typically folks are (re)using an RFID tag and scanning that rather than a printer barcode. I think it's OK without a backfill. Thanks for the review!

